PR #15, opened 2026-04-30
db(hardening): drop legacy migration_status + lock down public functions + tighten RLS

raphaelfh:
## Summary

Cleans up the security advisor gap inherited from the squashed `baseline_v1.sql`. After this PR the Supabase linter goes from **32 → 10** lints; every remaining lint is one of:
- a known operational table (`public.alembic_version` — needs no RLS)
- an intentional SECURITY DEFINER RPC that `authenticated` legitimately needs (`is_project_*`, `create_project_with_member`, etc.)
- a manual setting outside Alembic (`auth_leaked_password_protection` via dashboard, `pg_trgm` placement)

These migrations have already been applied to the remote project (`gdfslcfeobjdxihqtcsk`). Each is idempotent and ships with a downgrade path.

## Migrations

- **0007 — drop `migration_status`** ([0007_drop_migration_status.py](backend/alembic/versions/0007_drop_migration_status.py)). The table came in via the squashed `baseline_v1.sql` and is not read or written by any application code.

- **0008 — function hardening** ([0008_function_hardening.py](backend/alembic/versions/0008_function_hardening.py)).
  - Pins `search_path = public, pg_catalog` on every `public` function (closes `function_search_path_mutable`).
  - Revokes `EXECUTE` from `anon` / `PUBLIC` on every `SECURITY DEFINER` function. `authenticated` retains EXECUTE only where it actually needs it (RPCs and RLS helpers).

- **0009 — tighten RLS** ([0009_tighten_rls_policies.py](backend/alembic/versions/0009_tighten_rls_policies.py)).
  - `article_authors`: drops the blanket ALL-true policy and adds per-command policies. UPDATE/DELETE now require the caller to share a project with the author through `article_author_links → articles → project_members`. INSERT stays open (no project context exists before the link is created).
  - `feedback_reports`: INSERT is now `authenticated`-only and forces `user_id = auth.uid()` so users cannot spoof each other's feedback.

- **0010 — lock down `handle_new_user`** ([0010_lock_handle_new_user.py](backend/alembic/versions/0010_lock_handle_new_user.py)). The trigger fires with the function's owner privileges, so signed-in users do not need EXECUTE — and exposing it via `/rest/v1/rpc/handle_new_user` lets a logged-in user call it with a forged `NEW` record.

## Test plan

- [x] `ruff check` + `ruff format --check` pass on all 4 migration files
- [x] `alembic upgrade head` applied cleanly against remote (now at `0010_lock_handle_new_user`)
- [x] `get_advisors` security: 32 → 10 lints
- [ ] Smoke test login flow on staging (verifies `is_project_member` still callable by `authenticated` for RLS)
- [ ] Smoke test feedback form (verifies INSERT works for own user, fails for spoofed `user_id`)
- [ ] Verify CI green

### DIFF
--- a/backend/.env.example
+++ b/backend/.env.example
@@ -36,4 +36,4 @@ ENCRYPTION_KEY="review_hub_default_key_change_me_in_production"
 # =================== LANGSMITH (OPCIONAL) ===================
 LANGCHAIN_TRACING_V2=false
 LANGCHAIN_API_KEY=""
-LANGCHAIN_PROJECT="review-hub"
+LANGCHAIN_PROJECT="prumo"
--- a/backend/alembic/versions/0007_drop_migration_status.py
+++ b/backend/alembic/versions/0007_drop_migration_status.py
@@ -1,0 +1,37 @@
+"""drop legacy migration_status table
+
+Revision ID: 0007_drop_migration_status
+Revises: 0006_article_text_blocks
+Create Date: 2026-04-29
+
+The ``migration_status`` table came in via the squashed ``baseline_v1.sql``
+(it was created by an earlier project's bespoke migration tracker, before
+the project standardised on Alembic). No application code reads or writes
+it, so it's pure dead weight and triggers a ``rls_disabled_in_public``
+advisor lint.
+"""
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "0007_drop_migration_status"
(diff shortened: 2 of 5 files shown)


raphaelfh:
Closing to keep the PR queue focused on the screening workstream (PR #34). DB hardening can be revisited as a separate effort — branch `cleanup/post-008-hardening` preserved.

PR #165, opened 2026-05-30
test(zotero): mock supabase client and drop stale duplicate test

raphaelfh:
## Problem

`frontend/test/services/zoteroImportService.test.ts` failed at module load with `Error: supabaseUrl is required`. `zoteroImportService.ts` imports `@/integrations/supabase/client` at the top level, which calls `createClient(SUPABASE_URL, …)` — and that throws when no Supabase env vars are set in the test environment.

A second, stale copy of the suite lived at `frontend/tests/services/zoteroImportService.test.ts` — a one-line re-export shim of the canonical `frontend/test/` file. Vitest's default include glob picked up both, so the same suite ran (and failed) twice.

## Fix

- Add the conventional `vi.mock('@/integrations/supabase/client', …)` stub to the canonical test, matching sibling service tests (e.g. `templateImportService.test.ts`). One mock covers the whole import chain — `@/integrations/api/client` pulls `supabase` from the same module.
- Remove the stale `frontend/tests/` re-export shim. It was the only file under `frontend/tests/`; the project standardises on `frontend/test/`.

Both broken files predate this branch (they were already committed on `dev`), so landing this here fixes the suite for every branch once it syncs with `dev`.

## Verification

```
npm run test:run -- zoteroImportService
# Test Files  1 passed (1)
#      Tests  2 passed (2)
```


raphaelfh:
Closing as superseded by `dev`: the run-vocabulary copy changes and the Zotero test updates this branch carries are already merged to dev (net diff vs dev for these files is empty), and the branch is ~6.7k lines behind dev (the feedback #164 + e2e #170 work landed after it was cut). No unique changes remain to merge. Reopen if I've missed something.

PR #203, opened 2026-06-05
fix: remove dead instance.status check in extraction progress

### LINKED ISSUE #100: bug(ArticleExtractionTable): instance.status does not exist on ExtractionInstance — "all completed" branch is permanently dead, progress never reaches 100%
## Summary

`ArticleExtractionTable.tsx:430` checks `instance.status === 'completed'` to short-circuit the progress calculation at 100%. The `status` property does not exist on `ExtractionInstance` (confirmed by the type definition and by `TS2339`). At runtime `instance.status` is always `undefined`, the condition is always `false`, and the function falls through to the count-based estimate — meaning extraction progress can never reach 100% via the "all instances completed" fast path, even when every instance truly is done.

## File & Lines

`frontend/components/extraction/ArticleExtractionTable.tsx:425-441`

```ts
const calculateExtractionProgress = (article: ArticleWithExtraction) => {
  if (article.instances.length === 0) return 0;

  // Check if all instances have status 'completed'
  const allCompleted = article.instances.every(instance => instance.status === 'completed');
  //                                                       ^^^^^^^^^^^^^^^^
  //  TS2339: Property 'status' does not exist on type 'ExtractionInstance'
  //  Runtime value: undefined → allCompleted is always false
  if (allCompleted && article.instances.length > 0) {
    return 100;   // ← never reached
  }

  // Fallback: count instances with at least one extracted value
  const instancesWithValues = article.instances.filter(instance =>
    article.extractedValues.some(value => value.instance_id === instance.id)
  ).length;

  return Math.round((instancesWithValues / article.instances.length) * 100);
};
```

`ExtractionInstance` type (`frontend/types/extraction.ts:137-150`) has no `status` column:

```ts
export interface ExtractionInstance {
  id: string; project_id: string; article_id: string; template_id: string;
  entity_type_id: string; parent_instance_id: string | null;
  label: string; sort_order: number; metadata: any;
  created_by: string; created_at: string; updated_at: string;
  // no 'status' field
}
```

## Repro

1. Open any project with fully-extracted articles in the article extraction table.
2. Observe that even when all fields are filled and a run is finalized, the progress column never shows 100% via the "all completed" shortcut — it is gated on the `extractedValues` count fallback instead.
3. `npx tsc -p tsconfig.app.json --noEmit 2>&1 | grep 'ArticleExtractionTable.*TS2339'` → `error TS2339: Property 'status' does not exist on type 'ExtractionInstance'.`

## Verification

```
$ npx tsc -p tsconfig.app.json --noEmit 2>&1 | grep ArticleExtractionTable
frontend/components/extraction/ArticleExtractionTable.tsx(430,71): error TS2339: Property 'status' does not exist on type 'ExtractionInstance'.

$ grep -n "status" frontend/types/extraction.ts | grep -i "interface ExtractionInstance" -A20
# 'status' does not appear in the ExtractionInstance block (lines 137-150)
```

## Severity

**med** — the progress percentage shown in the article extraction table is systematically incorrect: fully-completed articles show a count-based estimate ≤100% rather than 100%, since the fast-path check is dead code. Affects all users who rely on the progress column to gauge extraction completion.

---

_Auto-found by scheduled bug-hunt agent. DOW=6 WP=1 scope=frontend/components/extraction._

duongynhi000005-oss:
## Summary

The `calculateExtractionProgress` function in `ArticleExtractionTable.tsx` checks `instance.status === 'completed'` to short-circuit the progress calculation at 100%. The `ExtractionInstance` interface (defined at line 81) does **not** have a `status` property — the check is always `false` and the early-return is dead code.

This causes a TypeScript error (`TS2339: Property 'status' does not exist on type 'ExtractionInstance'`).

## Fix

Remove the broken `allCompleted` check and keep the working fallback logic that correctly computes progress based on extracted values.

## Changes
- `frontend/components/extraction/ArticleExtractionTable.tsx`: Remove 6 lines of dead code (the `allCompleted` check and early return)

Fixes #100

### DIFF
--- a/frontend/components/extraction/ArticleExtractionTable.tsx
+++ b/frontend/components/extraction/ArticleExtractionTable.tsx
@@ -445,12 +445,6 @@ export function ArticleExtractionTable({ projectId, templateId }: ArticleExtract
   const calculateExtractionProgress = (article: ArticleWithExtraction) => {
     if (article.instances.length === 0) return 0;
 
-      // Check if all instances have status 'completed'
-    const allCompleted = article.instances.every(instance => instance.status === 'completed');
-    if (allCompleted && article.instances.length > 0) {
-      return 100;
-    }
-
       // Count instances with at least one extracted value
     const instancesWithValues = article.instances.filter(instance =>
       article.extractedValues.some(value => value.instance_id === instance.id)


raphaelfh:
Thanks for catching this and for the clear write-up! 🙏

We folded your fix into #209 (now merged to `dev` + `main`), which removes the same dead `instance.status` fast-path **and** the sibling `TS2339` in the same file (the no-op `HeaderCheckbox` indeterminate effect). That cleared issue #100, so this PR is now superseded — closing it as redundant. Much appreciated.

PR #259, opened 2026-06-11
test(qa): stub supabase.auth in fullscreen test, silence false error toast

raphaelfh:
## Why

The supabase client mock in `frontend/test/QualityAssessmentFullScreen.test.tsx` stubbed `.from()` but not `.auth`, so every render crashed the background AI-suggestions load (`supabase.auth.getUser()` of `undefined`) — producing a console.error and an "Error loading suggestions" toast in all 8 tests. The noise nearly masked the real publish-toast failure during the React 19 upgrade (#252).

## What changed

- Added an `auth.getUser` stub returning the `{ data: { user }, error: null }` shape `aiSuggestionService.loadSuggestions` expects (frontend/services/aiSuggestionService.ts:153); the downstream `extraction_reviewer_states` query is already covered by the mock's default-table branch.
- The publish preflight test now also asserts the suggestions error toast does NOT fire, pinning the regression.

## Risk

Test-only change; no production code touched.

## Test plan

- `npx vitest run frontend/test/QualityAssessmentFullScreen.test.tsx` — 8/8, zero getUser/suggestions noise in output
- `npm run test:run` — 451/451
- `npm run typecheck` clean; `npm run lint` — 0 errors, 94 warnings (baseline)

### DIFF
--- a/frontend/test/QualityAssessmentFullScreen.test.tsx
+++ b/frontend/test/QualityAssessmentFullScreen.test.tsx
@@ -74,6 +74,16 @@ vi.mock("@/integrations/supabase/client", () => {
 
   return {
     supabase: {
+      // useAISuggestions → AISuggestionService.loadSuggestions resolves the
+      // current reviewer via supabase.auth.getUser(); without this stub every
+      // render surfaces an "Error loading suggestions" toast that drowns out
+      // the assertions below.
+      auth: {
+        getUser: async () => ({
+          data: { user: { id: "qa-test-reviewer-id" } },
+          error: null,
+        }),
+      },
       from: (table: string) => {
         if (table === "project_extraction_templates") {
           return makeQuery(PROBAST_TEMPLATE);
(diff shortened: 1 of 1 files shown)


raphaelfh:
## Claude review
**Verdict:** LGTM
**Scope:** Test-only fix — adds a `supabase.auth.getUser` stub to the fullscreen QA test mock so background AI-suggestion loads don't crash and emit a spurious error toast, then pins the regression with a negative assertion in the publish-preflight test.

### Findings

### Incident-class sweep

| Class | Status |
|---|---|
| BOLA | n/a — no production code |
| TOCTOU | n/a — no production code |
| Swallowed errors | ok — regression assertion explicitly guards the error-swallow path |
| ApiResponse envelope | n/a — test-only |
| TanStack cache | n/a — test-only |
| Schema drift | n/a — test-only |
| RLS | n/a — test-only |

### Notes (non-blocking)

- **`afterEach` uses `vi.restoreAllMocks()`** (line 185), which only restores `vi.spyOn` spies — it does not reset `vi.fn()` call history. The publish test compensates correctly with explicit `(toast.error as ReturnType<typeof vi.fn>).mockClear()` at line 289 before the assertion. No issue, but worth knowing if more tests are added that assert `toast.error` call counts.
- The negative assertion (`expect(toast.error).not.toHaveBeenCalledWith(…)`) is synchronous after a `waitFor`. Since the `auth.getUser` stub prevents the error condition entirely, this is correct — no timing gap to worry about.

---

PR #260, opened 2026-06-11
refactor(frontend): fix react-hooks compiler warnings in ui + shared components

raphaelfh:
## Why

eslint-plugin-react-hooks 7 surfaces the React Compiler rules; the repo carries 94 pre-existing warnings held at `warn` (see `eslint.config.js`). This is batch 1 of the burn-down: all 10 warnings under `frontend/components/ui/**` and `frontend/components/shared/**`. Repo-wide count drops 94 → 84, 0 errors.

## What changed

All fixes use compiler-friendly patterns (lazy state init, adjust-during-render, derive-in-render) instead of `eslint-disable`:

- **set-state-in-effect (5)** — `SelectWithOther`, `MultiSelectWithOther`, `carousel`, `resizable-panel`, `EntitySelectorComparison`: prop→state mirrors moved from effects to `useState` initializers + the [adjust-during-render pattern](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes); `EntitySelectorComparison` now derives the default selection (`picked ?? first ?? null`) instead of auto-selecting via effect.
- **immutability (2)** — `ComparisonCell`: `handleSave`/`handleCancel` declared before the effect that closes over them; `resizable-panel`: `onPointerUp` unregisters itself via a named function expression instead of referencing its own `const` binding.
- **refs (1)** — `useResizableTableColumns`: latest-width ref written in an effect, not during render.
- **static-components (1)** — `file-upload-progress`: `StatusIcon` hoisted to module scope with a `status` prop (was re-created per render → remount).
- **purity (1)** — `sidebar` skeleton: `Math.random()` width via `useState` initializer (fixed once per mount, as before).

## Risk

Low; pure frontend refactor, no data-path or API changes. Two intentional micro-improvements to be aware of:

- `resizable-panel.tsx`: the close animation now actually plays. The old effect set `isClosing` only after the `collapsed=true` render had already returned `null` (unmounting the `<aside>`), so the documented width/opacity transition could never run on the same node. The render-phase adjustment keeps the node mounted across the flip — this is the behavior the file header and `docs/superpowers/design-system/sidebar-and-panels.md` describe.
- Mirrored state now initializes from props on mount (no one-frame empty flash before the old effects ran).

## Test plan

- [x] `npm run lint` — 84 problems (0 errors), all 10 batch warnings gone, none new

## Out of scope

The remaining 84 warnings (hooks/, components/extraction/, pages/, etc.) — follow-up batches in progress.

### DIFF
--- a/frontend/components/shared/comparison/ComparisonCell.tsx
+++ b/frontend/components/shared/comparison/ComparisonCell.tsx
@@ -70,6 +70,18 @@ export function ComparisonCell({
   // ✅ NOVO: Detectar se é campo numérico
   const isNumberField = field?.field_type === 'number';
 
+  const handleSave = () => {
+    if (editValue !== value && onValueChange) {
+      onValueChange(editValue);
+    }
+    setIsEditing(false);
+  };
+
+  const handleCancel = () => {
+    setEditValue(value);
+    setIsEditing(false);
+  };
+
   // ✅ NOVO: Detectar clique fora do editor
   useEffect(() => {
     if (!isEditing) return;
@@ -101,18 +113,6 @@ export function ComparisonCell({
       document.removeEventListener('mousedown', handleClickOutside);
     };
   }, [isEditing, editValue, value, onValueChange, isSelectOpen]);
-
-  const handleSave = () => {
-    if (editValue !== value && onValueChange) {
-      onValueChange(editValue);
-    }
-    setIsEditing(false);
-  };
-
-  const handleCancel = () => {
-    setEditValue(value);
-    setIsEditing(false);
-  };
 
   const handleKeyDown = (e: React.KeyboardEvent) => {
     if (e.key === 'Enter' && !e.shiftKey) {
--- a/frontend/components/shared/comparison/EntitySelectorComparison.tsx
+++ b/frontend/components/shared/comparison/EntitySelectorComparison.tsx
@@ -16,7 +16,7 @@
  * | Number of preds    | 5       | 3       | 5       |
  */
 
-import {useCallback, useEffect, useMemo, useState} from 'react';
+import {useCallback, useMemo, useState} from 'react';
 import {Label} from '@/components/ui/label';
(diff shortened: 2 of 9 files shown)


raphaelfh:
## Claude review
**Verdict:** LGTM
**Scope:** Batch 1 of the react-hooks compiler-warning burn-down: eliminates all 10 warnings in `ui/` and `shared/` by applying adjust-during-render, lazy `useState` initializers, module-scope component hoisting, and effect-based ref updates — no `eslint-disable` suppressions.

### Findings

<details>
<summary>Spot-checks performed</summary>

- **carousel.tsx** — `onSelect(api)` was removed from the effect and replaced with direct `setCanScrollPrev`/`setCanScrollNext` calls in the adjust-during-render block. Verified `onSelect` (line 53–60) only calls those two setters, so removal is equivalent and slightly better (avoids the post-paint flash).
- **useResizableTableColumns.ts** — moving `columnWidthsRef.current = columnWidths` into a `useEffect([columnWidths])` introduces a sub-frame staleness window. For a drag interaction (next `mousemove` fires at the earliest on the next animation frame, after the effect runs) this is negligible. Accepted cost of React Compiler compliance.
- **resizable-panel.tsx** — traced through the `collapsed` false→true→false scenario; the adjust-during-render correctly keeps the `<aside>` node mounted during `CLOSE_ANIMATION_MS` and the cleanup cancels the timer if re-opened early. The described animation fix is real.
- **SelectWithOther / MultiSelectWithOther** — adjust-during-render uses `value !== prevValue` (reference equality), same semantics as the original `useEffect([value])` deps. Confirmed no infinite-loop risk; the lazy `useState` initializers eliminate the first-render empty-flash.
</details>

### Incident-class sweep
| Class | Status |
|---|---|
| BOLA | n/a — no API calls or ownership checks touched |
| Run-state TOCTOU | n/a — no backend state mutations |
| Swallowed errors | n/a |
| ApiResponse envelope drift | n/a |
| TanStack cache staleness | n/a |
| Schema drift | n/a |
| RLS | n/a |

---

PR #261, opened 2026-06-11
refactor(frontend): fix react-hooks compiler warnings in pages, contexts, pdf-viewer

raphaelfh:
## Why

Batch 2 of the react-hooks compiler lint burn-down (94 pre-existing warnings held at `warn` in `eslint.config.js`). This batch clears all 12 warnings under `frontend/pages/**`, `frontend/contexts/**`, and `frontend/pdf-viewer/**`. Together with #260 the repo count is 94 → 72.

## What changed

Compiler-friendly patterns only, no `eslint-disable`:

- **set-state-in-effect (8)** — URL→tab sync (`ProjectContext`, `UserSettings`), spinner reset on project switch (`ProjectView`), template-resolution reset + proposal hydration (`QualityAssessmentFullScreen`), store→input mirrors (`NavigationControls`, `usePageHandle`): all moved to the [adjust-during-render pattern](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes). `ProjectView`'s load-kickoff effect now starts the loaders from a microtask so their setState calls run in async callbacks.
- **immutability (2)** — `ProjectView`: the kickoff effect referenced `loadProject`/`loadArticles` before their declarations; effect moved below them.
- **preserve-manual-memoization (2)** — `handleReopen` in both full-screen pages used optional-chained deps (`session?.runId`, `finalizedRun?.id`), which the compiler can't preserve; hoisted to plain identifiers.
- **globals (1)** — pdf-viewer context test captures the store api via `renderHook` instead of reassigning an outer `let` during render.

## Risk

Low. The adjust-during-render conversions fire on the same transitions the old effects watched; state lands one commit earlier (no behavior-visible difference, removes a cascading render). `ProjectView`'s generation-bump ordering is unchanged — the bump still happens in the effect body before the loaders capture it.

## Test plan

- [x] `npm run lint` — batch directories clean; repo 82 warnings on this branch (94 − 12; batch 1 merges separately), 0 errors
- [x] `npm run typecheck` — clean
- [x] `npm run test:run` — 451/451 passed
- [x] `npm run build` — success

## Out of scope

Remaining warnings in `frontend/hooks/**` and `frontend/components/**` — follow-up batches.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #262, opened 2026-06-11
refactor(frontend): fix react-hooks compiler warnings in shared hooks

raphaelfh:
## Why

Batch 3 of the react-hooks compiler lint burn-down. Clears all 14 warnings in the top-level/shared/performance/zotero hooks under `frontend/hooks/`. With #260 and #261 the repo count goes 94 → 58.

## What changed

- **set-state-in-effect (10)** — three sub-patterns:
  - `use-mobile`: `useIsMobile`/`useIsNarrow` rebuilt on `useSyncExternalStore` — matchMedia is an external store, so this is the canonical shape (also returns the correct value on first render instead of `false`-then-flip).
  - `useNavigation` breadcrumbs: pure function of the route, now derived with `useMemo` (state + effect deleted).
  - Fetch-on-mount loaders (`useNotifications`, `useUserProfile`, `useProjectsList`, `useProjectSettings`, `useProjectMemberRole`, `useZoteroIntegration`, `useComparisonPermissions`, `useZoteroSyncStatus`): kickoff deferred one microtask so the loaders' setState calls happen in async callbacks; param-clear resets moved to adjust-during-render with initial state derived from params.
- **refs (1)** — `useKeyboardShortcuts`: latest-bindings ref written in an effect, not during render.
- **use-memo (2)** — `usePreserveScroll`: dep on the stable `selectors` array instead of a spread (callers pass a module constant); `useOptimizedCacheWithDeps` deleted — zero callers, and `useCallback(fn, deps)` with a forwarded dep list is unfixably compiler-incompatible.
- **preserve-manual-memoization (pre-empted)** — `useProjectMemberRole` dep `user?.id` hoisted to a plain identifier so fixing the effect doesn't surface a new warning.

## Risk

Low. The microtask deferral shifts loader kickoff by one microtask (imperceptible; the fetches were already async). `useZoteroSyncStatus`'s `tick` already guards on `active`, so cleanup-before-microtask is safe. Removing `useOptimizedCacheWithDeps` is dead-code deletion verified by grep.

## Test plan

- [x] `npm run lint` — batch files clean; 80 warnings on this branch (94 − 14), 0 errors

## Out of scope

Remaining warnings in `frontend/components/**` and `frontend/hooks/{extraction,hitl,qa,runs}/**` — follow-up batches.

### DIFF
--- a/frontend/hooks/performance/useOptimizedCache.ts
+++ b/frontend/hooks/performance/useOptimizedCache.ts
@@ -1,4 +1,4 @@
-import {useCallback, useEffect, useMemo, useState} from 'react';
+import {useCallback, useEffect, useState} from 'react';
 import {articleCache, globalCache} from '@/services/cacheService';
 
 /**
@@ -259,26 +259,4 @@ export function useOptimizedObjectCache<T extends Record<string, any>>({
     updateFields,
     getField,
   };
-}
-
-/**
- * Hook para cache de dados com dependências
- */
-export function useOptimizedCacheWithDeps<T>(
-  cacheKey: string,
-  fetchFn: () => Promise<T>,
-  deps: React.DependencyList,
-  options: Omit<UseOptimizedCacheOptions<T>, 'cacheKey' | 'fetchFn'> = {}
-) {
-  const memoizedFetchFn = useCallback(fetchFn, deps);
-  const memoizedCacheKey = useMemo(() => 
-    `${cacheKey}_${JSON.stringify(deps)}`, 
-    [cacheKey, deps]
-  );
-
-  return useOptimizedCache({
-    cacheKey: memoizedCacheKey,
-    fetchFn: memoizedFetchFn,
-    ...options,
-  });
 }
--- a/frontend/hooks/shared/useComparisonPermissions.ts
+++ b/frontend/hooks/shared/useComparisonPermissions.ts
@@ -57,9 +57,19 @@ export function useComparisonPermissions(
     canManageBlindMode: false,
     canExport: false,
     canEditTemplate: false,
-    loading: true,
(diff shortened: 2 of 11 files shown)


raphaelfh:
## Claude review
**Verdict:** LGTM
**Scope:** Batch 3 of React compiler lint burn-down — 14 warnings cleared across shared/performance/zotero hooks using `useSyncExternalStore`, adjust-during-render, `useMemo` derivation, latest-bindings ref fix, and microtask deferral.

### Findings

Spot-checks worth noting (all came up clean):
- `useMediaQuery` (`use-mobile.tsx`): `useSyncExternalStore` without `getServerSnapshot` is fine for a Vite SPA — no SSR hydration concern.
- `useKeyboardShortcuts.ts`: 1-render stale window after moving `bindingsRef.current` write to `useEffect` is inconsequential — keyboard events only fire after paint.
- `usePreserveScroll.ts`: `[selectors]` vs `[...selectors]` — the pre-existing comment already documents the stable-reference contract; behaviorally correct given callers pass module-level constant arrays.
- `useOptimizedCacheWithDeps` deletion: dead-code removal, PR body references grep verification.
- Microtask deferral throughout: React 18 silently no-ops setState after unmount, so the unmount-before-microtask window is safe.

### Incident-class sweep

| Class | Status |
|---|---|
| TOCTOU (run-state) | n/a — no run-state transitions |
| Swallowed errors | ok — all loaders retain their own try/catch; `void` prefix on microtask calls is intentional |
| ApiResponse envelope | n/a — no fetch layer changes |
| TanStack cache | n/a — these hooks use direct useState, not TanStack Query |
| RLS | n/a — no migrations |

---

PR #263, opened 2026-06-11
refactor(frontend): fix react-hooks compiler warnings in feature components

raphaelfh:
## Why

Batch 4 of the react-hooks compiler lint burn-down. Clears all 16 warnings under `frontend/components/{articles,project,user,hitl,navigation}` (one extra surfaced in `ProfileSection` once the file became compilable — fixed here too). With #260–#262 the repo count goes 94 → 42.

## What changed

- **incompatible-library (3)** — `AddProjectDialog`, `ProfileSection`, `SecuritySection`: `form.watch(...)` → `useWatch({control, name})`, react-hook-form's compiler-compatible subscription API. Same values, same re-render triggers.
- **set-state-in-effect (5)** — dialog open/param resets (`ArticlesExportDialog`, `ZoteroImportDialog`, `HITLArticleTable`) moved to adjust-during-render; consensus drafts (`ReviewConsensusSection`, `TemplateConsensusOverride`) now initialize from the TanStack cache and re-hydrate during render — the mount case is covered by the lazy initializer, so cached data no longer flashes defaults for one frame.
- **immutability (7)** — effects that referenced loaders declared below them (`ArticleDetailDialog`, `ArticleForm`, `TeamMembersSection`, `ApiKeysSection`, `ProfileSection`): effects moved below the loader declarations, kickoff deferred one microtask so the loaders' setState calls run in async callbacks.
- **purity (2)** — `Date.now()` reads hoisted to module-scope helpers: `buildStorageKey` (AddArticleDialog upload key) and `countRecentlyFinished` (NotificationCenter unread badge). Memoization semantics unchanged — the unread count still refreshes only when the job list changes, as before.

## Risk

Low; UI-layer refactor, no data-path changes. The consensus-draft hydration is the only logic worth a second look: previously hydration always happened in a post-mount effect; now cached data hydrates in the initializer and later refetches hydrate during render — same data, one commit earlier ([ReviewConsensusSection.tsx:62](frontend/components/project/settings/ReviewConsensusSection.tsx)).

## Test plan

- [x] `npm run lint` — batch directories clean; 78 warnings on this branch (94 − 16), 0 errors

## Out of scope

Remaining warnings in `frontend/components/extraction/**` and `frontend/hooks/{extraction,hitl,qa,runs}/**` — final two batches.


raphaelfh:
## Claude review
**Verdict:** LGTM
**Scope:** Frontend-only React Compiler lint burn-down (batch 4): 16 warnings cleared across `articles`, `project`, `user`, `hitl`, and `navigation` components using four techniques — `useWatch`, adjust-during-render, lazy `useState` initializers, and `queueMicrotask`-deferred effects.

### Findings

The one place worth a second read (called out in the PR body) is the consensus-draft hydration in `ReviewConsensusSection.tsx:62` and `TemplateConsensusOverride.tsx:47`. The pattern is correct: the lazy initializer handles the first render from cache; the `prevConfigData !== projectConfig.data` guard re-hydrates on every subsequent refetch. Both components' draft state correctly tracks the server config through project-ID changes and cache updates. No stale-draft risk above the old effect-based approach.

`countRecentlyFinished` in `NotificationCenter.tsx:38`: `Date.now()` is called at invocation time (inside `useMemo`), identical semantics to the old inline version — refreshes only when `recentJobs` changes, no regression.

`queueMicrotask` pattern: each loader handles its own errors internally (try/catch), so `void` is appropriate and nothing is swallowed.

### Incident-class sweep

| Class | Status |
|---|---|
| Swallowed errors | ok — `void` + loader-internal try/catch throughout |
| TanStack cache | ok — lazy initializer + during-render re-hydration is correct; cache keys unchanged |
| Schema drift | n/a — no type or model changes |
| RLS | n/a — no migrations |

---

PR #264, opened 2026-06-11
refactor(frontend): fix react-hooks compiler warnings in extraction components

raphaelfh:
## Why

Batch 5 of the react-hooks compiler lint burn-down. Clears all 20 warnings under `frontend/components/extraction/**` (one extra surfaced in `EditFieldDialog` once the file became compilable). With #260–#263 the repo count goes 94 → 22.

## What changed

- **incompatible-library (3)** — `AddFieldDialog`, `AddSectionDialog`, `EditFieldDialog`: `form.watch(...)` → `useWatch({control, name})` (incl. the inline `watch('allow_other')` in EditFieldDialog's JSX).
- **set-state-in-effect (11)** — three sub-patterns, consistent with the earlier batches:
  - dialog open/close resets and selection syncs → adjust-during-render (`ExtractionExportDialog`, `ImportTemplateDialog`, `RemoveSectionDialog`, `AddModelDialog`, `AISuggestionHistoryPopover`, `ExtractionInterface` URL/template syncs);
  - "auto-select first item" → render-phase invariants with terminating guards (`ModelLevelComparison`, `ExtractionExportDialog` reviewer default, `ExtractionInterface` non-manager tab clamp);
  - async loaders kicked off from microtasks (`ExtractionInterface`, `TemplateConfigEditor`, `ArticleExtractionTable`, `AISuggestionHistoryPopover`, `RemoveSectionDialog`, `EditFieldDialog`).
- **immutability (5)** — TDZ effect/loader ordering fixes, plus one real fix: `InstanceCard` was mutating the `instance` prop in place after a label save ([InstanceCard.tsx:75](frontend/components/extraction/InstanceCard.tsx) before); the saved label now lives in local state, re-synced if the prop changes from outside. Rendered output is identical.
- **static-components (1)** — `ArticleExtractionTable`'s `HeaderCheckbox` was a `React.memo` created *inside* the component body — a new component type every render, which defeats the memo and remounts the node. Hoisted to module scope (it captured nothing).

## Risk

Low-moderate: this is the extraction surface, so the dialogs and table were the focus of the test pass. The `InstanceCard` change is the only semantic candidate — the prop mutation never triggered re-renders elsewhere, so replacing it with card-local state is observably equivalent.

## Test plan

- [x] `npm run lint` — extraction directory clean; 74 warnings on this branch (94 − 20), 0 errors
- [x] `npm run typecheck` — clean
- [x] `npm run test:run` — 451/451 passed
- [x] `npm run build` — success

[...]

Issue #23, opened 2026-05-02
bug(hooks): useExtractionSession cancelledRef reset race — stale POST response overwrites new session


raphaelfh:
## Summary

`useExtractionSession` uses `cancelledRef` to prevent stale in-flight responses from updating state after the component unmounts or props change. However the ref is **reset to `false` inside the new effect before the old in-flight request can observe `true`**, so the guard never fires on a prop change and the stale session from the previous article/project overwrites the correct one.

## File & Lines

`frontend/hooks/extraction/useExtractionSession.ts:91–97`

```ts
useEffect(() => {
  cancelledRef.current = false;   // ← Step 2: resets the flag immediately
  void open();
  return () => {
    cancelledRef.current = true;  // ← Step 1: cleanup fires first (React guarantees)
  };
}, [open]);
```

React runs cleanup **then** the new effect body synchronously, before any pending microtask/promise resolves. So the sequence when `projectTemplateId` changes is:

1. Cleanup: `cancelledRef.current = true`
2. New effect: `cancelledRef.current = false`   ← cancellation signal immediately cleared
3. Previous in-flight `POST /api/v1/hitl/sessions` resolves → sees `false` → calls `setSession()` with stale `run_id`

The new `open()` then also resolves and calls `setSession()` with the correct `run_id`, but there is a window where `runId` from the old article is the active session and autosave may write to it.

## Repro (3 lines)

1. Navigate quickly between two articles (A → B) while the HITL session request for A is still in-flight.
2. Article B's effect resets `cancelledRef.current` to `false`.
3. Article A's response resolves, sets `session.runId` to A's run ID; any autosave fired in that window posts decisions to the **wrong run**.

## Verification

Code inspection: `cancelledRef.current = false` at line 92 executes synchronously during the new effect, before any pending `Promise` (the old `open()` fetch) can settle. The `if (!cancelledRef.current)` check at line 71 therefore never sees `true` for a prop-change scenario — only for a genuine unmount where no new effect fires.

## Severity

**high** — fast article navigation can route autosave writes to the previous article's extraction run. The user sees no error (writes succeed), but data lands in the wrong run and may be overwritten or lost when the stale session is eventually replaced.

## Fix hint

Use a per-call unique token (or `AbortController`) instead of a shared ref:

```ts
useEffect(() => {
  let cancelled = false;
  const run = async () => {
    // ... fetch ...
    if (!cancelled) setSession(...);
  };
  void run();
[...]

PR #265, opened 2026-06-11
refactor(frontend): fix react-hooks compiler warnings in domain hooks

raphaelfh:
## Why

Final batch (6/6) of the react-hooks compiler lint burn-down: all 22 warnings under `frontend/hooks/{extraction,hitl,qa,runs}` — the HITL/extraction data-path hooks. Together with #260–#264 this takes the repo from 94 warnings to **0**.

## What changed

- **`useAutoSaveProposals` (8 warnings)** — the riskiest file, please read first:
  - latest-prop ref mirrors (`valuesRef` etc.) written in an effect declared before their consumers;
  - the `'dirty'` badge flips via adjust-during-render keyed by values **content** (`JSON.stringify`) — identity-keyed tracking re-renders forever when a caller rebuilds the map每render (the hook's own test harness does);
  - `hasUnsavedChanges` is now a `useMemo` over a new `lastSavedByKey` **state** mirror (updated when saves acknowledge) instead of reading the mutable ref during render.
- **Session hooks (`useExtractionSession`, `useQAAssessmentSession`)** — `open()` split into a render-phase loader reset + `openCore` with no setState before its first `await`. The POST still starts synchronously inside the effect — the stale-generation tests (#23/#109) encode that ordering — via the same ref indirection `useModelManagement` already used. Manual `refetch` keeps its synchronous spinner.
- **Fetch-on-mount/param-change loaders** (`useAISuggestions`, `useExtractedValues`, `useExtractionData`, `useFieldManagement`, `useFinalizedExtractionRun`, `useGlobalTemplates`, `useHITLProjectTemplates`, both colaboracao hooks) — param-clear resets adjusted during render with lazy initial loading state; loader kickoff deferred one microtask so their setState calls run in async callbacks.
- **`useModelManagement`** — only the render-phase ref write moved into an effect; the synchronous mount load is untouched (`createModel`'s optimistic ordering depends on it).
- **`useExtractionData`** — `template?.id` dep hoisted to a plain identifier (preserve-manual-memoization).

## Risk

Moderate — this is the autosave/session heart of the HITL stack. Mitigations: the existing test suite is dense here (41 tests across the three riskiest hooks, including the stale-generation and state-machine contracts) and caught two intermediate regressions during development (microtask-deferred POST start; identity-keyed render loop), both fixed by restoring the original synchronous semantics. Generation-bump ordering in the session hooks is byte-identical.

## Test plan

- [x] `npm run lint` — `hooks/{extraction,hitl,qa,runs}` clean; 72 warnings on this branch (94 − 22), 0 errors
- [x] `npm run test:run` — 451/451 passed (including useAutoSaveProposals state-machine + useExtractionSession #23 stale-response tests)

## Out of scope

Nothing — this completes the 94-warning burn-down. Once all six PRs merge, the `warn` downgrades in `eslint.config.js` can be promoted back to `error` (follow-up).

### DIFF
--- a/frontend/hooks/extraction/ai/useAISuggestions.ts
+++ b/frontend/hooks/extraction/ai/useAISuggestions.ts
@@ -119,7 +119,8 @@ export function useAISuggestions(props: UseAISuggestionsProps): UseAISuggestions
     // useEffect AFTER loadSuggestions declaration
   useEffect(() => {
     if (!enabled || !articleId) return;
-    loadSuggestions();
+    // Microtask so the loader's setState calls run in an async callback.
+    queueMicrotask(() => void loadSuggestions());
   }, [articleId, enabled, loadSuggestions]);
 
   const acceptSuggestionCore = useCallback(async (instanceId: string, fieldId: string, silent: boolean): Promise<boolean> => {
--- a/frontend/hooks/extraction/colaboracao/useAllUserInstances.ts
+++ b/frontend/hooks/extraction/colaboracao/useAllUserInstances.ts
@@ -30,21 +30,17 @@ export function useAllUserInstances(props: UseAllUserInstancesProps): UseAllUser
   const { articleId, enabled = true } = props;
 
   const [instances, setInstances] = useState<InstanceWithCreator[]>([]);
-  const [loading, setLoading] = useState(true);
+  // Only show the loader when there is actually something to load.
+  const [loading, setLoading] = useState(() => Boolean(enabled && articleId));
   const [error, setError] = useState<string | null>(null);
   const generationRef = useRef(0);
 
-  useEffect(() => {
-    if (!enabled || !articleId) {
-      setLoading(false);
-      return;
-    }
-
-    loadInstances();
-    return () => {
-      generationRef.current += 1;
-    };
-  }, [articleId, enabled]);
+  // Params cleared after mount: stop the loader (during render, not via effect).
+  const [prevKey, setPrevKey] = useState({ articleId, enabled });
+  if (articleId !== prevKey.articleId || enabled !== prevKey.enabled) {
+    setPrevKey({ articleId, enabled });
+    if (!enabled || !articleId) setLoading(false);
+  }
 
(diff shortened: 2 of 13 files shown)


raphaelfh:
## Claude review
**Verdict:** LGTM with nits
**Scope:** Fixes 22 React Hooks compiler warnings across 13 HITL/extraction domain hooks (the final batch of a 6-PR burn-down from 94→0 warnings) by applying three patterns: `queueMicrotask` to defer setState calls in effects; adjust-during-render (ADR) for loading/error resets on param change; effect-based ref syncing instead of render-phase writes.

### Findings

- **nit** `frontend/hooks/extraction/useExtractionSession.ts:69–95` / `frontend/hooks/qa/useQAAssessmentSession.ts:70–100` — The ADR sets `setLoading(true)` when `willOpen` is truthy, but there is no reset path in the ADR when params subsequently transition `willOpen → false` before the effect fires. In that edge case `loading` stays `true` until the next successful `openCore` or unmount. The original code never set `loading=true` unless the guard passed inside `open()`, so this is a behavioural delta. Low impact in practice because callers gate rendering on `enabled`, but worth being aware of. Suggested fix: add `else { setLoading(false); }` to the `if (willOpen)` branch in the ADR block (or confirm callers never expose the stuck state).

### Incident-class sweep

| Class | Status |
|---|---|
| BOLA | n/a — no backend endpoints changed |
| TOCTOU (run-state) | ok — generation-bump ordering preserved byte-for-byte; `openCoreRef` sync effect declared before the call effect in both session hooks, maintaining source-order guarantee |
| Swallowed errors | ok — all catch blocks and `Promise.allSettled` usage in `useAutoSaveProposals` preserved; `useHITLProjectTemplates` error handling inside IIFE retained |
| ApiResponse envelope | n/a — no changes to API call sites |
| TanStack cache | n/a — `queryClient.invalidateQueries` calls untouched |
| Schema drift | n/a — pure frontend hooks |
| RLS | n/a — no migrations |

---

PR #266, opened 2026-06-11
refactor(extraction)!: typed LLM call layer on Pydantic AI + Logfire (full migration)

raphaelfh:
## Summary

Full migration of the extraction LLM stack from the hand-rolled OpenAI-over-httpx wrapper to a typed call layer, per the approved spec ([docs/superpowers/specs/2026-06-11-extraction-llm-stack-design.md](docs/superpowers/specs/2026-06-11-extraction-llm-stack-design.md)) and plan ([docs/superpowers/plans/2026-06-11-extraction-llm-stack-migration.md](docs/superpowers/plans/2026-06-11-extraction-llm-stack-migration.md)).

**New `backend/app/llm/` package — the single doorway to LLMs:**
- `provider.py` — BYOK key resolution → `OpenAIChatModel` (Anthropic later = one new branch); explicit `MissingLLMKeyError` replaces the silent `Bearer None`
- `schema.py` — runtime Pydantic models from `extraction_fields` rows: real types, `Literal` enums from `allowed_values`, alias-safe field names, required-field hints, OpenAI strict-mode chunker (~14 fields/call)
- `prompts/` — versioned templates (`NAME` + content-hash `VERSION` stamped on every span)
- `validators.py` — evidence plausibility via `ModelRetry` semantic reask
- `extractor.py` — `extract_structured`: NativeOutput, reask budget (2), `UsageLimits` ceiling, `LlmUsage` keeping the legacy `tokens_*` vocabulary
- `observability.py` — Logfire wiring (FastAPI lifespan, Celery `worker_process_init` post-fork, structlog correlation), fully inert without `LOGFIRE_TOKEN`

**Both services migrated** (section + model identification): same public APIs, same DB writes, same `tokens_*` JSONB keys, same structlog events, same API envelopes — the frontend does not notice.

**Behavior changes (intended):**
- Invalid LLM output no longer degrades to a silent empty dict — it reasks up to 2x then fails the run with a clear error
- Mid-batch LLM failures now fail only that section's run (`fail_run`) instead of rolling back the whole uncommitted batch transaction — fixes a latent data-loss path that the new raise-on-invalid semantics would have made mainstream
- Zero-field templates skip the LLM call entirely

**Demolition (no legacy left):** `openai_service.py` (446 LOC), `json_parser.py` (248 LOC), `app/services/llm/`, and the dead `langchain`/`langchain-openai`/`instructor`/`tenacity` deps — 29 packages out of the lock. Canonical refs (`extraction-hitl-architecture.md` §4.2, `constitution.md`) updated.

## Verification

- 675 backend unit tests green (37 new in `tests/unit/llm/`, incl. 1 opt-in live smoke via `PRUMO_LLM_SMOKE=1`); suite hard-blocks real LLM calls (`ALLOW_MODEL_REQUESTS=False`)
- `verify_all.sh` (ruff, eslint, tsc, pytest incl. integration, vitest, fitness) green at every phase gate and at the end
- Every task passed independent spec-compliance + code-quality subagent review; final whole-diff review verdict: ready to merge

## Deploy note

Set `LOGFIRE_TOKEN` on both Railway services (web + worker) to activate LLM observability — everything stays inert without it.


raphaelfh:
## Claude review

Full diff read (8 518 lines, 45 files). Auth, schema drift, ApiResponse envelope, run-state concurrency, migrations, and TanStack cache are **clean** for this change. The main risk surface is the new LLM error-handling layer: `MissingLLMKeyError` is caught correctly on the batch path but leaks to the generic handler on two other paths.

---

### Critical

**C1 — `MissingLLMKeyError` falls through to `rollback_and_fail` in standalone `extract_section`**

`backend/app/services/section_extraction_service.py` — the batch/memory path explicitly catches `(AgentRunError, MissingLLMKeyError)` and routes them to `fail_run`. The standalone `extract_section` path wraps `_extract_with_llm` in a single generic `except Exception` that calls `rollback_and_fail` for everything. A missing BYOK key in the standalone path will therefore roll back a committed transaction instead of just marking the run failed.

Fix: add `except (AgentRunError, MissingLLMKeyError)` before the generic handler in the standalone path, mirroring the batch split.

**C2 — Same gap in `model_extraction_service.extract()`**

`backend/app/services/model_extraction_service.py` — `_identify_models` calls `build_model(model, api_key=self._llm_api_key)` with no guard. The outer `extract()` try-except (which wraps the entire flow) has no `MissingLLMKeyError` branch; the exception hits the generic `rollback_and_fail`. Same consequence as C1 for the model-identification path.

---

### Important

**I1 — Zero-field template skip silently produces no-op suggestions**

`backend/app/services/section_extraction_service.py` — when `build_output_models` returns an empty list, `_extract_with_llm` returns `({}, LlmUsage())` and logs `extraction_skipped_no_fields`. The callers that receive the empty dict do not check for this condition before creating suggestions. An empty dict is indistinguishable from a successful extraction with all-null fields at the call site. At minimum the callers should gate on the `extraction_skipped_no_fields` case; otherwise an empty-template misconfiguration silently creates zero-valued suggestion rows.

**I2 — Chunked extraction merge does not validate completeness**

`backend/app/services/section_extraction_service.py` — the loop that merges per-chunk results calls `extracted_data.update(dump_extraction(output))` with no assertion that the union of chunk fields covers the full entity-type field list. If one chunk returns an empty payload (silent failure upstream), those fields disappear from the merged dict without any log warning. Downstream JSONB writes will have missing keys with no trace of the gap.

**I3 — Prompt version hash is import-time; long-lived Celery workers won't see hot updates**

`backend/app/llm/prompts/__init__.py` — `content_version()` hashes the prompt text when the module is first imported. Celery workers are persistent across deployments unless explicitly restarted. A rolling deploy that ships a new prompt template will stamp the *old* version hash on every span until the worker restarts. The version pinning is useful precisely for debugging, so a stale hash is worse than no hash. Add a note in the deploy checklist (or `observability.py`) that workers must restart on prompt changes, or compute the hash lazily per-call.

**I4 — `UsageLimits(request_limit=5)` ceiling is untested**

`backend/app/llm/extractor.py:121` — the PR description calls this a key circuit-breaker for BYOK cost control, but no test exercises the limit actually stopping a call. The `test_live_extraction_round_trip` opt-in smoke test does not mock an agent that exceeds the budget. Add a unit test that injects an agent mock returning `UsageLimitExceeded` and verifies the exception propagates (rather than being swallowed by a future catch block).

---

### Minor

**M1 — No test for `MissingLLMKeyError` in standalone section-extraction or model-extraction paths**

`TestExtractSectionLlmFailure` covers `UnexpectedModelBehavior` and reask exhaustion but not a missing key. Given C1 and C2 above, adding a test here would both catch the regression and document the expected routing (`fail_run` vs `rollback_and_fail`).

**M2 — Worker observability bootstrap has no exception guard**

`backend/app/worker/celery_app.py` — `_configure_worker_observability` is a `@worker_process_init.connect` handler with no try-except. If `logfire.configure()` raises (e.g., bad token, import error), the worker child crashes on startup with no structured log. Wrapping in a broad `except Exception` + `structlog.get_logger().warning(...)` would keep the worker alive and observable.

**M3 — PDF text truncation at 15 000 chars may split multi-byte UTF-8**

`backend/app/llm/prompts/__init__.py:217` — `text[:MAX_PDF_CHARS]` slices by character index, which is correct in Python (characters, not bytes). However, if the upstream extraction step ever passes raw bytes decoded with a lossy codec, a surrogate character could land at position 15 000 and produce an invalid string. Low probability, but a `text[:MAX_PDF_CHARS].encode("utf-8", errors="replace").decode("utf-8")` round-trip is a one-liner guard. Existing tests use ASCII only, so this is untested.

---

### Clean

- **Authorization (BOLA):** No new endpoints; existing `ensure_project_member` calls untouched.
- **RLS / multi-tenancy:** No new tables or policy changes.
- **Run-state TOCTOU:** All transitions go through `run_lifecycle_service`; no ad-hoc `run.status =` assignments found.
- **Error swallowing (structural):** No `.catch(() => success)` patterns; `ModelRetry` raises correctly; `Promise.all` vs `allSettled` not applicable (Python).
- **Schema drift:** `LlmUsage` and `Evidence` use `ConfigDict(extra="forbid")`; no `Optional` / nullable mismatches detected.
- **ApiResponse envelope:** No new HTTP endpoints; internal service contract unchanged.
- **Migrations:** None added; Alembic untouched.
- **Conventional commits + PR targets `dev`:** ✓

---

Blocking on C1 and C2 (error-handling gap that changes run fate on missing key). I1 and I2 are worth addressing before merge given they're silent data-quality paths. The rest can be follow-up issues if the author prefers.

`pr_review_done status=needs-changes findings=critical:2,important:4,minor:3`

---

PR #267, opened 2026-06-11
chore(lint): promote react-hooks compiler rules from warn back to error

raphaelfh:
## Why

Closes out the react-hooks compiler lint burn-down. The nine `react-hooks/*` rules were temporarily downgraded to `warn` while the 94 pre-existing findings were fixed; with #260–#265 all merged, `eslint .` is clean repo-wide, so the rules return to their recommended `error` level. From here, any new violation fails CI instead of accumulating as debt.

## What changed

- `eslint.config.js`: removed the nine per-rule `warn` overrides (the rules now inherit `error` from `reactHooks.configs.recommended.rules`) and replaced the debt comment. `exhaustive-deps` stays `off` and the non-react-hooks rules are untouched.

## Risk

None at runtime — config-only. The only failure mode is a latent warning somewhere lint doesn't currently see, and `eslint .` exits 0 on this branch.

## Test plan

- [x] `npm run lint` — exits clean, 0 problems, with rules at error level

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -21,22 +21,12 @@ export default tseslint.config(
       "react-refresh": reactRefresh,
     },
     rules: {
+      // React Compiler rules from eslint-plugin-react-hooks 7 run at their
+      // recommended error level. The 94-finding debt that once held them at
+      // "warn" was burned down in PRs #260–#265 — keep them at error so new
+      // violations fail CI instead of accumulating.
       ...reactHooks.configs.recommended.rules,
         "react-hooks/exhaustive-deps": "off",
-        // eslint-plugin-react-hooks 7 ships the React Compiler rules at
-        // error level. 94 pre-existing findings (52x set-state-in-effect,
-        // 16x immutability, 9x refs, ...) are real but predate the gate —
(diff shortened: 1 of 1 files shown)


raphaelfh:
## Claude review
**Verdict:** LGTM
**Scope:** Config-only promotion of 9 `react-hooks` compiler rules from `warn` back to `error` in `eslint.config.js`, after the 94-finding pre-existing violation debt was cleared in PRs #260–#265.

### Findings

### Incident-class sweep

| Class | Status |
|---|---|
| BOLA | n/a — no endpoints |
| Swallowed errors | n/a — no async code |
| ApiResponse envelope | n/a — no API changes |
| TanStack cache | n/a — no query changes |
| Schema drift | n/a — no model/type changes |
| RLS | n/a — no migrations |

The spread order is correct: `...reactHooks.configs.recommended.rules` sets the 9 compiler rules to `error`, then `"react-hooks/exhaustive-deps": "off"` overrides just that one rule — intentional and accurate. Commit follows Conventional Commits, PR targets `dev`, test plan is complete.


---

PR #291, opened 2026-06-14
fix(backend): deterministic id tiebreak for AI-proposal superseded detection

raphaelfh:
## Problem

The AI-metadata export's proposal query in `extraction_export_service._load_ai_proposal_rows` ordered only by `created_at`, so the **latest per `(run, instance, field)`** — which drives the FR-037 `superseded` `reviewer_outcome` — was **nondeterministic** when two `source='ai'` proposals share a `created_at`. Same-transaction inserts share PostgreSQL's transaction-start `now()`, so this is reachable in practice, and the `superseded` label could flip between export builds.

## Fix

Add `ExtractionProposalRecord.id.desc()` as a secondary sort key, matching the canonical `ExtractionProposalRepository.get_latest_for_coord` ordering (which documents `id` as the deterministic tiebreaker on equal `created_at`). One line, no new abstraction — it just aligns this bulk query with the repo's established pattern.

## Test

`tests/integration/test_extraction_export_ai_outcome_ordering.py` inserts two `source='ai'` proposals for the same coord with an identical explicit `created_at` and hand-picked min/max UUIDs, then asserts the higher-id row is `latest` (→ `pending`, no decisions) and the lower-id is `superseded`, **stable across repeated builds**. It exercises the real DB query ordering via `_load_ai_proposal_rows` (not a reimplementation).

The prior unit determinism test only exercised pre-built `AIProposalRow` objects, so the DB query ordering was uncovered.

## Verification

- **Red→green proven:** with the tiebreak reverted, the test fails on all 3 random seeds (1, 42, 109494657); restored, it passes on all 3.
- 130 existing export unit tests pass; `ruff check` + `ruff format --check` clean.

### DIFF
--- a/backend/app/services/extraction_export_service.py
+++ b/backend/app/services/extraction_export_service.py
@@ -1189,7 +1189,11 @@ class ExtractionExportService(LoggerMixin):
         }
 
         # 1. AI proposals, newest first so the latest is encountered before
-        # any superseded ones for the same (run, instance, field).
+        # any superseded ones for the same (run, instance, field). ``id`` is
+        # the deterministic tiebreaker on equal ``created_at`` (same-transaction
+        # inserts share the timestamp), matching the canonical
+        # ExtractionProposalRepository.get_latest_for_coord ordering so the
+        # FR-037 "superseded" outcome is stable across export builds.
         proposal_rows = (
             await self.db.execute(
                 select(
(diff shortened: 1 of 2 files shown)


raphaelfh:
## Claude review
**Verdict:** LGTM
**Scope:** Adds `id.desc()` as a secondary sort key in `_load_ai_proposal_rows` so the "superseded" `reviewer_outcome` on the AI metadata export is deterministic when two `source='ai'` proposals share a `created_at`; covers with an integration test that forces the tie.

### Findings

No findings above the confidence bar — the deterministic CI gates cover the rest.

### Incident-class sweep

| Class | Status |
|---|---|
| BOLA | n/a — no endpoint touched |
| TOCTOU | n/a — read-only query reorder |
| Swallowed errors | n/a |
| TanStack cache | n/a — backend only |
| Schema drift | n/a — no model/schema changes |
| RLS | n/a — no new tables |

Fix is a one-liner that matches the established `get_latest_for_coord` ordering (`extraction_proposal_repository.py:74-77`). The integration test correctly forces the equal-timestamp scenario with hand-picked UUIDs and exercises `_load_ai_proposal_rows` against the real DB rather than mocking the query result. The `pytest.skip` guard is a safe dead branch — the autouse `SEED` fixture in `tests/integration/conftest.py` always seeds extraction instances before the suite runs.

---

PR #292, opened 2026-06-15
feat(export): publication-ready xlsx export redesign (design, plan + phases S1–S2…)

raphaelfh:
Rebuilds the extraction `.xlsx` export into publication-ready supplemental material (template-agnostic). This PR lands the **design spec**, the **implementation plan**, and **Phase S1 — the envelope-aware value resolver**, which fixes the root cause of the "data is wrong / AI-metadata is buggy" reports.

## Docs
- `docs/superpowers/specs/2026-06-14-publication-ready-xlsx-export-design.md` — redesign spec (supersedes shipped 009).
- `docs/superpowers/plans/2026-06-14-publication-ready-xlsx-export.md` — 76-task, 8-phase TDD plan.

## Phase S1 — value resolver (7 commits)
The shipped export unwrapped only single-key `{value}` envelopes, so `{value,unit}`, `{selected,other_text}` / `{selected:[…],other_texts:[…]}`, and double-wrapped values leaked into cells as Python-repr dict strings (`{'value': 5, 'unit': 'mg'}`) across the **matrix, per-reviewer columns, and the AI-metadata sheet**.

- `resolve_value` — one pure, no-IO envelope resolver; recursively unwraps, surfaces `unit` (`5 mg`), resolves `other` free text.
- Routed **all** value-maps (consensus / single-user / all-users) **and** both AI value columns through it.
- Shared format helper across matrix + AI cells; `_xlsx_safe` now **raises** on a stray dict instead of silently `str()`-ing it; `_unwrap_value` deleted.

## Verification
- 154 export tests green (unit + integration); the value-resolution integration test proven **red→green** (old code leaks `{'value':5,'unit':'mg'}`, new yields `"5 mg"`).
- ruff check + format + layered-arch fitness clean. Every task passed spec-compliance **and** code-quality review.

## Notes
- First of 8 phases; **S2–S8 stack on this branch**.
- The AI-proposal `id`-tiebreak (spec §6.2 A6) ships separately in #291 — this PR intentionally leaves that proposal-ordering line untouched.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #300, opened 2026-06-17
fix(frontend): surface async-export completion so the file actually downloads

raphaelfh:
## Problem

After the publication-ready xlsx export redesign (#292) shipped to prod, a real export (mode=all_users, 19 articles, AI metadata) **built and signed the file on the worker** (Railway worker log: `task_completed … succeeded in 2.2s`, 25,961-byte `.xlsx` uploaded + signed URL returned), and the status endpoint returned `completed` with the `download_url` — yet the user reported the export **never downloads**.

## Root cause

`useBackgroundJobPolling` only iterated `getActiveJobs()`, which the store defines as **`running`/`pending` only** (`useBackgroundJobs.ts`). A job is therefore invisible the instant it becomes `completed`, so the `running→completed` transition was never observed and **`onJobComplete` / `onJobFailed` could never fire** — they were dead code. Result: no completion toast and no download action for async exports.

This is a latent bug in the shared async-job pattern (introduced with articles-export); the export redesign exposed it because a large/all-users/AI-metadata export is forced down the async path.

## Fix

Subscribe to the full `jobs` list and diff each job's previous status, firing the terminal-state callbacks exactly once on the `running→completed` (and `→failed`) transition. `updateJob` always produces a new array, so the completion reliably re-runs the effect. Removed the now-dead `interval` option.

## Tests

- New `frontend/hooks/useBackgroundJobPolling.test.tsx` — RED before the fix (callbacks fired 0×), GREEN after. Covers completion, failure, and "don't re-toast an already-terminal job on mount".
- Full suite green: **515 tests / 80 files**, `typecheck` clean, `eslint` clean (incl. React Compiler rules).

### DIFF
--- a/frontend/components/navigation/NotificationCenter.tsx
+++ b/frontend/components/navigation/NotificationCenter.tsx
@@ -142,9 +142,8 @@ export function NotificationCenter() {
         };
     }, [updateJob]);
 
-    // Polling to update jobs
+    // Observe job status transitions to surface completion/failure toasts.
   useBackgroundJobPolling({
-    interval: 2000,
     onJobComplete: (job) => {
       toast.success(getCompletionMessage(job), {
         duration: 5000,
--- a/frontend/hooks/useBackgroundJobPolling.test.tsx
+++ b/frontend/hooks/useBackgroundJobPolling.test.tsx
@@ -1,0 +1,105 @@
+/**
+ * Regression coverage for useBackgroundJobPolling.
+ *
+ * Root-cause guard for the async-export "never downloads" incident
+ * (extraction + articles export): the completion observer must fire
+ * `onJobComplete` / `onJobFailed` when a background job transitions
+ * INTO a terminal state. The original implementation only inspected
+ * `getActiveJobs()` (running|pending), so a job became invisible the
+ * instant it completed and the completion toast (+ download action)
+ * never appeared — the worker built and signed the file, but the user
+ * never saw a download.
+ */
+
+import {beforeEach, describe, expect, it, vi} from "vitest";
+import {act, renderHook} from "@testing-library/react";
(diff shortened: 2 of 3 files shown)


raphaelfh:
## Claude review
**Verdict:** LGTM with nits
**Scope:** Replaces the interval-based `useBackgroundJobPolling` implementation — which polled `getActiveJobs()` (running|pending only) and therefore could never observe a `running→completed` transition — with a Zustand subscription to the full `jobs` list, diffing each job's previous status in a `useEffect` to fire `onJobComplete`/`onJobFailed` exactly once on a terminal transition.

### Findings

- **nit** `frontend/hooks/useBackgroundJobPolling.ts:40` — `onJobComplete` and `onJobFailed` are in the `useEffect` dependency array. In `NotificationCenter.tsx` both are defined as inline arrow functions, so they get new references on every render and the effect re-runs more than necessary. The `previousStatusRef` guard prevents any double-firing, so this is harmless, but it creates unnecessary churn. Suggested fix: wrap both callbacks in `useCallback` in `NotificationCenter.tsx`, or if the intent is to keep this hook ignorant of the caller, use a `useRef`-based stable callback pattern inside the hook itself (capture the latest callback in a ref and call the ref inside the effect). Either approach is fine; the current code is correct but slightly wasteful.

### Incident-class sweep

| Class | Status |
|---|---|
| BOLA | n/a — pure frontend, no API endpoints |
| TOCTOU (run-state) | n/a — no backend state mutations |
| Error swallowing | ok — optional chaining (`?.`) on callbacks, no suppression |
| ApiResponse envelope | n/a — no API calls in this diff |
| TanStack cache | n/a — Zustand store, not TanStack Query |
| Schema drift | n/a — no backend or type changes |
| RLS | n/a — no migrations |

---

PR #301, opened 2026-06-17
Promote to main: async-export completion download fix (#300)

raphaelfh:
Promotes the verified fix from `dev` to `main` (production deploy via Railway + Vercel).

### Included
- **#300** fix(frontend): surface async-export completion so the file actually downloads.
  `useBackgroundJobPolling` only iterated `getActiveJobs()` (running|pending), so the `running→completed` transition was never observed and `onJobComplete`/`onJobFailed` were dead code — async extraction/articles exports built + signed the file on the worker but no completion toast/download ever appeared. Now subscribes to the full job list and diffs each job's previous status.

### Verification
- New `useBackgroundJobPolling.test.tsx` (RED→GREEN), full frontend suite **515 tests green**, typecheck + eslint clean.
- Root cause corroborated by Railway worker logs (worker `task_completed`, 25,961-byte `.xlsx` signed) + status endpoint returning `completed` while the UI never surfaced it.

### DIFF
--- a/frontend/components/navigation/NotificationCenter.tsx
+++ b/frontend/components/navigation/NotificationCenter.tsx
@@ -142,9 +142,8 @@ export function NotificationCenter() {
         };
     }, [updateJob]);
 
-    // Polling to update jobs
+    // Observe job status transitions to surface completion/failure toasts.
   useBackgroundJobPolling({
-    interval: 2000,
     onJobComplete: (job) => {
       toast.success(getCompletionMessage(job), {
         duration: 5000,
--- a/frontend/hooks/useBackgroundJobPolling.test.tsx
+++ b/frontend/hooks/useBackgroundJobPolling.test.tsx
@@ -1,0 +1,105 @@
+/**
+ * Regression coverage for useBackgroundJobPolling.
+ *
+ * Root-cause guard for the async-export "never downloads" incident
+ * (extraction + articles export): the completion observer must fire
+ * `onJobComplete` / `onJobFailed` when a background job transitions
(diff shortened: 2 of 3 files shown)


raphaelfh:
## Claude review
**Verdict:** LGTM with nits
**Scope:** Converts `useBackgroundJobPolling` from `setInterval`-based polling of `getActiveJobs()` (running|pending only) to a Zustand subscription over the full `jobs` list, making the `running→completed` transition observable; adds regression tests that were RED under the old implementation.

### Findings

- **nit** `frontend/components/navigation/NotificationCenter.tsx` — `onJobComplete` and `onJobFailed` are passed as inline arrow functions, so they carry new references on every `NotificationCenter` render. The effect's dependency array includes both callbacks (`[jobs, onJobComplete, onJobFailed]`), so the effect body iterates all jobs on every parent render even when `jobs` hasn't changed. No correctness risk — the `seen` map guard (`previous === job.status`) prevents spurious callbacks — but it's unnecessary work. Fix: wrap both callbacks in `useCallback` in `NotificationCenter`.

### Incident-class sweep

| Class | Status |
|---|---|
| BOLA | n/a — frontend-only, no API endpoints |
| TOCTOU | n/a — no backend state transitions |
| Swallowed errors | ok — no catch blocks, no silent failures introduced |
| ApiResponse envelope | n/a — no API calls in changed code |
| TanStack cache | n/a — no TanStack Query usage |
| Schema drift | n/a — no type or schema changes |
| RLS | n/a — no migrations |

---

PR #302, opened 2026-06-17
fix(frontend): clear empty-state for the export reviewer picker

raphaelfh:
## Problem

On a project with no reviewer decisions yet (e.g. a freshly-imported project), the export dialog's **Single user** reviewer dropdown is empty, but the dialog still defaulted the selection to the current user (not in the list) and kept **Export** enabled. Result: a blank dropdown the user can't resolve, and submitting only returns `422 EMPTY_ELIGIBLE_ARTICLES`. Reported from a real project ("sfa") that has 19 articles but 0 instances / runs / reviewer decisions.

The picker query itself is correct — it returns reviewers-with-decisions and works on projects that have data (verified against another project where it correctly lists the reviewer).

## Fix

- Reconcile the reviewer selection against the eligible list: default to "me" only when I'm eligible; clear a selection that isn't in the list so Export stays blocked until a real reviewer is chosen.
- When **no** reviewer is eligible, show a clear empty-state message in place of the dropdown and disable Export (and hide the misleading "Will export N articles" preview).
- `modeReady` now requires an eligible selection in single-user mode.

Picker behaves normally on projects that have reviewer decisions.

## Tests

- New `ExtractionExportDialog.test.tsx` case: empty reviewers → empty-state shown, picker hidden, Export disabled (RED→GREEN).
- Full suite green: **516 tests / 80 files**, typecheck + eslint clean.

### DIFF
--- a/frontend/components/extraction/ExtractionExportDialog.test.tsx
+++ b/frontend/components/extraction/ExtractionExportDialog.test.tsx
@@ -264,6 +264,25 @@ describe("ExtractionExportDialog", () => {
         );
     });
 
+    it("blocks Single-user export with an empty-state when no reviewer has decisions", async () => {
+        useEligibleReviewersMock.mockReturnValue({data: [], isLoading: false});
+        const user = userEvent.setup();
+        renderDialog();
+        await user.click(screen.getByLabelText(/Single user/i));
+
+        // Empty-state message replaces the picker — there is nobody to pick.
+        await waitFor(() =>
+            expect(
+                screen.getByTestId("extraction-export-reviewer-empty"),
+            ).toBeInTheDocument(),
+        );
+        expect(
+            screen.queryByTestId("extraction-export-reviewer-picker"),
+        ).not.toBeInTheDocument();
+        // Export is blocked so the user cannot submit a doomed request.
+        expect(screen.getByTestId("extraction-export-submit")).toBeDisabled();
+    });
+
     it("renders the anonymize-reviewers toggle only when All-users mode + manager (US3 / FR-028)", async () => {
         const user = userEvent.setup();
         renderDialog();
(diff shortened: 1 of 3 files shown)


raphaelfh:
## Claude review
**Verdict:** LGTM
**Scope:** Fixes the export dialog defaulting to an ineligible reviewer (causing a blank, unresolvable dropdown) by reconciling the selection against the eligible list and showing a clear empty-state when no reviewer has decisions on the template.

### Findings

Two things I checked closely and found sound:

- **Render-phase reconciliation termination** (`ExtractionExportDialog.tsx:162–168`): when `userId` is not in the eligible list and `self` is `undefined`, `setReviewerId(null)` is called with the state already `null` — React's `Object.is` bailout prevents an infinite loop. ✓
- **Test validity** (`ExtractionExportDialog.test.tsx:265–283`): `data-testid="extraction-export-reviewer-picker"` exists on line 358 of the component (the manager `<Select>`), so the assertion `queryByTestId("extraction-export-reviewer-picker") not.toBeInTheDocument()` is a meaningful check, not a vacuous pass. ✓

One implicit behaviour change worth knowing: `modeReady` now also returns `false` while the reviewers query is still loading (since `reviewers.some(…)` over `[]` is always false). This is strictly safer than before (the old `Boolean(reviewerId)` would have allowed submit during loading), but the Export button is now momentarily disabled on dialog open until the query resolves. Acceptable trade-off.

### Incident-class sweep

| Class | Status |
|---|---|
| BOLA | n/a — no new endpoints; backend 422 is unchanged |
| TOCTOU | n/a — no run-state mutations |
| Swallowed errors | ok |
| TanStack cache | n/a — no new queries or mutations |
| Schema drift | n/a — frontend-only |
| RLS | n/a — no DB changes |

---

PR #303, opened 2026-06-17
Promote to main: export reviewer empty-state fix (#302)

raphaelfh:
Promotes the reviewer-picker empty-state fix from `dev` to `main` (production).

- **#302** fix(frontend): clear empty-state for the export reviewer picker. When a project has no eligible reviewers, the dialog now shows a clear empty-state and blocks Export instead of rendering a blank dropdown with a phantom selection.

Verified: dialog test (RED→GREEN), full suite 516 green, typecheck + eslint clean.

### DIFF
--- a/frontend/components/extraction/ExtractionExportDialog.test.tsx
+++ b/frontend/components/extraction/ExtractionExportDialog.test.tsx
@@ -264,6 +264,25 @@ describe("ExtractionExportDialog", () => {
         );
     });
 
+    it("blocks Single-user export with an empty-state when no reviewer has decisions", async () => {
+        useEligibleReviewersMock.mockReturnValue({data: [], isLoading: false});
+        const user = userEvent.setup();
+        renderDialog();
+        await user.click(screen.getByLabelText(/Single user/i));
+
+        // Empty-state message replaces the picker — there is nobody to pick.
+        await waitFor(() =>
+            expect(
+                screen.getByTestId("extraction-export-reviewer-empty"),
+            ).toBeInTheDocument(),
+        );
+        expect(
+            screen.queryByTestId("extraction-export-reviewer-picker"),
+        ).not.toBeInTheDocument();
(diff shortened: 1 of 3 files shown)


raphaelfh:
## Claude review
**Verdict:** LGTM with nits
**Scope:** Adds an empty-state guard to `ExtractionExportDialog` — when `useEligibleReviewers` returns an empty list in single-user mode the picker is replaced by an explanatory paragraph and Export is blocked; the `modeReady` predicate is tightened to validate the selected reviewer is actually in the eligible list rather than trusting any non-null value.

### Findings

- **nit** `frontend/components/extraction/ExtractionExportDialog.tsx:149` — query error state is conflated with the empty state. When `reviewersQuery` fails (network error, 5xx), `isLoading` is `false` and `data` is `undefined`, so `reviewers = []` and `noEligibleReviewers = true`. The user then sees *"No reviewer has recorded decisions on this template yet…"* — a data-state message that is factually wrong; the real problem is a fetch error. Export is correctly blocked in both cases, so this is UX-only, not a data-integrity issue. Fix: gate on `!reviewersQuery.isError` as well and render a distinct error state (or at minimum preserve the original message for the data-empty path only).

  ```tsx
  const noEligibleReviewers =
      mode === "single_user" &&
      !reviewersLoading &&
      !reviewersQuery.isError &&   // ← add this
      reviewers.length === 0;
  ```

### Incident-class sweep

| Class | Status |
|---|---|
| BOLA | n/a — frontend-only, no new endpoints or authorization surface |
| TOCTOU | ok — `modeReady` now validates reviewer against the live eligible list; backend is authoritative at submit time |
| Swallowed errors | finding — see above (query error → wrong empty-state message) |
| TanStack cache | ok — existing hook consumed, no new mutations or invalidations |
| Schema drift | n/a — no backend/Pydantic/migration changes |
| RLS | n/a — no schema or policy changes |

---

Issue #55, opened 2026-05-09
bug(hooks): useExtractionProgress overstates completion — counts unique field types, not (instance × field) pairs


raphaelfh:
## Summary

`useExtractionProgress` (the local-state progress hook used on `ExtractionFullScreen`) builds `requiredFields` as a flat list of unique field definitions (one entry per field, regardless of how many instances exist), then marks a field "completed" if **any** instance in the local `values` map has a non-empty value for that `fieldId`. For templates with `cardinality='many'` entity types (e.g. multiple prediction models), filling a required field in **one** instance causes `completionPercentage` to reach 100 % while the remaining instances are still empty.

## File & Lines

`frontend/hooks/extraction/useExtractionProgress.ts:35–70`

```ts
// requiredFields always has instanceIds: [] — the TODO was never completed
const requiredFields: Array<{ fieldId: string; instanceIds: string[] }> = [];
entityTypes.forEach(entityType => {
  entityType.fields
    .filter(field => field.is_required)
    .forEach(field => {
      requiredFields.push({
        fieldId: field.id,
        instanceIds: [] // ← "Will be filled with real instances" — never is
      });
    });
});

// completedRequired checks ANY instance, not ALL instances
const completedRequired = requiredFields.filter(({ fieldId }) => {
  const hasValue = Object.keys(values).some(key => {
    const [, fId] = key.split('_');
    return fId === fieldId && values[key] !== null && values[key] !== undefined && values[key] !== '';
  });
  return hasValue;           // ← true as soon as ONE instance has a value
}).length;
```

## Repro

1. Create a project template with a `cardinality='many'` entity type (e.g. "Prediction Model") that has at least one required field (e.g. `model_type`).
2. Open `ExtractionFullScreen` for an article with 3 model instances; fill `model_type` for Model 1 only, leaving Models 2 and 3 empty.
3. Observe `completionPercentage = 100`, `isComplete = true`, and the "Submit for review" button enabled — despite two-thirds of the required work being undone.

## Verification

Code inspection: `requiredFields` at line 35 has `instanceIds: []` with a comment "Will be filled with real instances" that was never honoured. The `completedRequired` filter at line 51 uses `.some(key => ...)` across all `values` keys, so the first hit (any instance) terminates the search. `totalFields` equals unique field-definition count, not `instances × fields`.

[...]

PR #304, opened 2026-06-17
fix(extraction): stop optional entities blocking the finalize gate

raphaelfh:
## Problem

The extraction completion percentage gates the "Submit for review" button (`isComplete` = 100%). It injected a mandatory **phantom instance** for *every* entity type with required fields but zero instances — ignoring the template's own `is_required` flag.

For the CHARMS template, `prediction_models` is **optional** (`is_required=false`, `cardinality=many`) with 6 child sections (Model Performance alone has 35 required fields). An article with all **33** study-level required fields filled but **no models added** read **33 / 82 = 40%** and could never reach 100%, so the button stayed disabled with no explanation.

The backend has **no** completeness gate for `advance_stage` PROPOSAL→REVIEW ([run_lifecycle_service.py](backend/app/services/run_lifecycle_service.py)), so this was a purely frontend block, stricter than the actual contract. The same wrong number also showed in the HITL article list, extraction table and dashboard (shared `computeRowProgress`).

## Fix (template-driven — not CHARMS-specific)

When the authoritative instance set is supplied, a zero-instance entity type contributes to the denominator **only if the template marks it `is_required`**:

- optional entity (and its child sections) with no instances → contributes nothing → zero models = complete;
- required entity with no instances → keeps a phantom slot → form stays below 100% until one is added.

No entity names are hardcoded — it keys off `cardinality` / `is_required` / instance existence, so it adapts to each user's template. The header path now passes the materialized `instances` so it gets the same authoritative behaviour as the tables. The value-key fallback keeps the historical phantom-1 for callers that omit instances (regression #55 preserved).

## Files

- `frontend/lib/extraction/progress.ts` — `is_required`-aware denominator
- `frontend/hooks/extraction/useExtractionProgress.ts` — accepts + forwards `instances`
- `frontend/pages/ExtractionFullScreen.tsx` — passes `instances` into the hook

## Verification

- **Live prod run** (`263f4b75…`, the reported one): post-fix math = **33 / 33 = 100%** → button enables. No data change needed; resolves on deploy.
- **TDD:** added tests that fail red then pass green — the exact `40 → 100` for a no-model article, the required-entity-stays-gated case, and the header path; regression #55 still passes.
- 365 extraction-area tests pass; `tsc --noEmit` clean; eslint clean.

### DIFF
--- a/frontend/hooks/extraction/useExtractionProgress.ts
+++ b/frontend/hooks/extraction/useExtractionProgress.ts
@@ -19,9 +19,33 @@ import {
 
 export type UseExtractionProgressReturn = RequiredFieldProgress;
 
+/** Minimal projection of a materialized instance the metric needs. */
+export interface ProgressInstanceProjection {
+  id: string;
+  entity_type_id: string;
+}
+
 export function useExtractionProgress(
   values: Record<string, any>,
   entityTypes: ProgressEntityProjection[],
+  instances?: ProgressInstanceProjection[],
 ): UseExtractionProgressReturn {
-  return computeRequiredFieldProgress(values, entityTypes);
+  // When the materialized instances are known, pass the TRUE instance set so the
+  // denominator reflects reality: an optional cardinality='many' entity (e.g.
+  // "Prediction Models" with zero models added) and its child sections add no
+  // required-field slots, so a fully-filled form is not stranded below 100%.
+  // Without instances (e.g. mid-load) we fall back to the value-key heuristic.
+  let instanceIdsByEntityType: Map<string, Set<string>> | undefined;
+  if (instances && instances.length > 0) {
+    instanceIdsByEntityType = new Map<string, Set<string>>();
+    for (const inst of instances) {
+      let set = instanceIdsByEntityType.get(inst.entity_type_id);
+      if (!set) {
+        set = new Set();
+        instanceIdsByEntityType.set(inst.entity_type_id, set);
+      }
+      set.add(inst.id);
+    }
+  }
+  return computeRequiredFieldProgress(values, entityTypes, instanceIdsByEntityType);
 }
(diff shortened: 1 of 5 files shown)


raphaelfh:
## Claude review
**Verdict:** LGTM with nits
**Scope:** Fixes the "40%, can't submit" bug by making `computeRequiredFieldProgress` skip optional `cardinality=many` entity types (and their child sections) when an authoritative instance set is passed and those entity types have zero instances — template-driven via `is_required`, no hardcoded names. The header path is wired up by passing `instances` into `useExtractionProgress`; `computeRowProgress` already received its instances, so it picks up the fix automatically.

### Findings

- **nit** `frontend/pages/ExtractionFullScreen.tsx:292` — The extended comment retains its Portuguese prefix (`// Hook para calcular progresso. Pass the materialized instances…`). CLAUDE.md §1 requires English-only comments; since the PR already touches this line, suggest: `// Progress hook — pass the materialized instances so optional …` (drop the Portuguese prefix entirely).

### Incident-class sweep

| Class | Status |
|---|---|
| BOLA | n/a — frontend-only; no endpoint changes |
| TOCTOU | n/a — pure synchronous computation, no state-machine writes |
| Error swallowing | ok — no async or catch paths introduced |
| ApiResponse envelope | n/a — no API calls changed |
| TanStack cache | n/a — no queries or mutations touched |
| Schema drift | ok — `is_required: boolean` is present on `ExtractionEntityTypeWithFields` (from `frontend/types/extraction.ts:101`) and satisfies `ProgressEntityProjection`; the authoritative path correctly uses it |
| RLS | n/a — no migrations |

#### Additional notes
- The `instances.length > 0` guard in `useExtractionProgress` treats an empty array as "not yet loaded", falling back to the value-key heuristic. This is safe: a zero-instance state always shows 0% regardless of path, and in practice the backend creates the singleton immediately on session open. The comment documents this intent clearly.
- Test coverage is solid: the exact `40 → 100` scenario, the required-entity-stays-gated case, the header path, and the pre-existing regression (#55) all covered. `tsc --noEmit` and eslint clean per the PR body.


---

PR #306, opened 2026-06-17
test(e2e): accept empty-state in single-user reviewer control + de-flake

raphaelfh:
## Root cause (systematic-debugging)

`Frontend E2E (ephemeral stack)` hard-failed on `extraction-export.e2e.ts › Single-user mode reveals the reviewer picker` (failed initial + retry):

```
expect(received).toContain(expected)
Expected value: null
Received array: ["picker", "locked"]
> expect(["picker", "locked"]).toContain(visible);
```

`ExtractionExportDialog` renders **three** mutually exclusive controls in single-user mode — `extraction-export-reviewer-picker` (manager), `-locked` (reviewer), and **`-empty`** (no eligible reviewers). The `-empty` branch was added by #302 ("clear empty-state for the export reviewer picker") and the **unit** test was updated, but the **E2E** test was not — it still asserted only picker-or-locked. In the ephemeral stack the export article has no eligible reviewers (`useEligibleReviewers` → `[]`), so the component correctly shows `-empty` and the assertion fails.

Compounding flake: the test used `Promise.race` over two non-retrying `isVisible()` checks, so it resolved to whichever settled first — returning `null` even when an element was visible.

Unrelated to the finalize-gate fix (#304/#305) — that change touches only the completion metric and appears nowhere in this spec; #304 carried the identical E2E result and merged (non-required check).

## Fix

Assert an auto-retrying `.or()` locator over all three testids. The test now verifies the real invariant (single-user mode reveals *a* reviewer control) and no longer depends on seed reviewer data or check timing.

## Verification

- `npm run typecheck` (CI `tsconfig.app.json`, includes `frontend/e2e`) — clean
- `eslint` on the file — clean
- `playwright test --list` — test collected under the new title
- Full ephemeral E2E runs in CI on this PR (cannot spin the stack locally)

Note: tests 276/300 (`extraction-export-button` visibility) flaked separately and passed on retry #1 — a softer timing flake, left out of scope.

### DIFF
--- a/frontend/e2e/flows/extraction-export.e2e.ts
+++ b/frontend/e2e/flows/extraction-export.e2e.ts
@@ -325,7 +325,7 @@ test.describe("Extraction export — UI flow", () => {
     ).toBeVisible();
   });
 
-  test("Single-user mode reveals the reviewer picker", async ({ page }) => {
+  test("Single-user mode reveals the reviewer control", async ({ page }) => {
     const env = loadE2EEnv();
     await loginViaUi(page);
     await page.goto(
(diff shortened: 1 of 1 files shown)


raphaelfh:
## Claude review
**Verdict:** LGTM
**Scope:** Fixes a broken E2E assertion in the single-user reviewer-control test by replacing a racy `Promise.race(isVisible, isVisible)` with Playwright's auto-retrying `.or()` locator chain, and adds the third testid (`-empty`) that was introduced by #302 but never covered in this test.

### Findings

### Incident-class sweep

| Class | Status |
|---|---|
| BOLA | n/a — test file only, no endpoint changes |
| TOCTOU | n/a |
| Swallowed errors | ok — the old `Promise.race` that could resolve to `null` (false-negative) is gone |
| TanStack cache | n/a |
| Schema drift | n/a |
| RLS | n/a |

---

**Notes:**

- All three testids (`-picker`, `-locked`, `-empty`) confirmed present in `ExtractionExportDialog.tsx:347,358,376`. The `.or()` chain matches the component's actual branching logic.
- The `Promise.race` bug was subtle: both `isVisible()` calls resolved immediately (they are non-retrying single-shot checks), so the race resolved to whichever settled first — returning `null` whenever the rendered element's check "lost", causing the containing `expect([...]).toContain(null)` to fail. The `.or()` locator is idiomatic Playwright and auto-retries with the default timeout, eliminating the race entirely.
- Test title rename ("picker" → "control") correctly tracks what the test now asserts.

---

PR #307, opened 2026-06-17
fix(frontend): notification bell reflects in-session jobs (the real 'never downloads' fix)

raphaelfh:
## Root cause (the actual "never downloads")

After [#300](https://github.com/raphaelfh/prumo/pull/300)/[#302](https://github.com/raphaelfh/prumo/pull/302) shipped, driving a real async export on prod showed the backend completes (worker builds + signs the `.xlsx`, status returns `completed` + `download_url`) but the **NotificationCenter never reflected the job in-session** — no badge, no list entry, no clickable download. Only a full page reload surfaced it.

`NotificationCenter` line 61 was:
```ts
const recentJobs = useMemo(() => getRecentJobs(20), [jobs, getRecentJobs]);
```
The callback never references `jobs` — `getRecentJobs` reads the store via the zustand getter. The **React Compiler derives deps from the callback body, not the manual array**, so it memoized `recentJobs` on the stable `getRecentJobs` action and never recomputed when `jobs` changed. (The `// kept:` comment shows the author tried to force recompute via a manual dep — the compiler doesn't honor it.) Store + poller were correct; only the rendered list was stale → bell showed jobs present at mount and missed in-session additions.

Confirmed on prod: store/localStorage had 4 jobs, bell rendered 3; a reload reconciled it.

## Fix

Extract a pure `selectRecentJobs(jobs, limit)` and derive `recentJobs` from the **reactive `jobs`** value so the compiler tracks the dependency. The bell now updates the instant a job is added/finishes.

## Tests

- New `NotificationCenter.test.tsx`: a job added **after mount** must appear without a reload — **RED before, GREEN after**, run under the React Compiler (vitest uses the same compiler config, so it reproduces the staleness).
- Full suite green: **521 tests / 81 files**, typecheck + eslint clean.

Combined with [#300](https://github.com/raphaelfh/prumo/pull/300) (completion observer) and [#302](https://github.com/raphaelfh/prumo/pull/302) (empty picker), this makes async extraction exports reach the user.

### DIFF
--- a/frontend/components/navigation/NotificationCenter.tsx
+++ b/frontend/components/navigation/NotificationCenter.tsx
@@ -20,7 +20,7 @@ import {
 } from '@/components/ui/dropdown-menu';
 import {ScrollArea} from '@/components/ui/scroll-area';
 import {Progress} from '@/components/ui/progress';
-import {useBackgroundJobs} from '@/stores/useBackgroundJobs';
+import {selectRecentJobs, useBackgroundJobs} from '@/stores/useBackgroundJobs';
 import {useBackgroundJobPolling} from '@/hooks/useBackgroundJobPolling';
 import {cn} from '@/lib/utils';
 import {toast} from 'sonner';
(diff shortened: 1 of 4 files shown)


raphaelfh:
## Claude review
**Verdict:** LGTM
**Scope:** Fixes a React Compiler memoization stale-read in `NotificationCenter` — extracts `selectRecentJobs` as a pure function so the compiler can derive a real `jobs` dependency from the callback body, replacing the old `useMemo(() => getRecentJobs(20), [jobs, getRecentJobs])` whose body never read `jobs` and was therefore memoized as stale. Also adds a regression test and fixes a flaky Playwright `Promise.race` to use the retrying `.or()` locator.

### Findings

### Incident-class sweep

| Class | Status |
|---|---|
| BOLA | n/a — frontend-only, no endpoints touched |
| TOCTOU | n/a — no backend state machine |
| Swallowed errors | ok — test mocks resolve, async assertions use `findByText` |
| ApiResponse envelope | n/a — no API call changes |
| TanStack cache | n/a — Zustand store, not TanStack |
| Schema drift | n/a — no type/schema changes |
| RLS | n/a — no migrations |

---

PR #308, opened 2026-06-17
Promote to main: notification bell reactivity fix (#307)

raphaelfh:
Promotes the real "never downloads" fix from `dev` to `main` (production).

- **#307** fix(frontend): notification bell reflects in-session jobs. `recentJobs` was memoized stale under the React Compiler (manual `[jobs]` dep not honored because the callback never referenced `jobs`), so async export jobs only appeared in the bell after a reload. Now derived from the reactive `jobs` via a pure `selectRecentJobs`.

Verified: NotificationCenter RED→GREEN test under the compiler; full suite 521 green; typecheck + eslint clean.

### DIFF
--- a/frontend/components/navigation/NotificationCenter.tsx
+++ b/frontend/components/navigation/NotificationCenter.tsx
@@ -20,7 +20,7 @@ import {
 } from '@/components/ui/dropdown-menu';
 import {ScrollArea} from '@/components/ui/scroll-area';
 import {Progress} from '@/components/ui/progress';
-import {useBackgroundJobs} from '@/stores/useBackgroundJobs';
+import {selectRecentJobs, useBackgroundJobs} from '@/stores/useBackgroundJobs';
 import {useBackgroundJobPolling} from '@/hooks/useBackgroundJobPolling';
 import {cn} from '@/lib/utils';
 import {toast} from 'sonner';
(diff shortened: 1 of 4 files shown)


raphaelfh:
## Claude review
**Verdict:** LGTM
**Scope:** Extracts a pure `selectRecentJobs` selector from the Zustand store so `useMemo` in `NotificationCenter` references `jobs` in its callback body — making the React Compiler's dep-tracking correct and fixing the bell's stale-on-mount behavior for in-session jobs.

### Findings

One informational note (pre-existing, not introduced here): `selectRecentJobs` uses `Math.max(limit - activeJobs.length, MAX_COMPLETED_JOBS)` for the finished-job slice, which can exceed `limit` when there are many active jobs (e.g. 15 active + 10 finished = 25 with `limit=20`). This is identical to the original `getRecentJobs` logic and appears intentional — the constant acts as a floor. No action needed.

### Incident-class sweep

| Class | Status |
|---|---|
| BOLA | n/a — frontend-only, no endpoints touched |
| TOCTOU | n/a — no backend state machines |
| Swallowed errors | ok — pure function, no async paths |
| ApiResponse envelope | n/a — no API calls in the diff |
| TanStack cache | n/a — Zustand store, no TanStack changes |
| RLS | n/a — no migrations, no Supabase changes |

**Fix correctness:** The React Compiler derives dependencies from the callback body, not the manual dep array. `useMemo(() => getRecentJobs(20), [jobs, getRecentJobs])` never referenced `jobs` inside the callback, so the compiler memoized on the stable `getRecentJobs` action and never recomputed on new jobs. The fix (`selectRecentJobs(jobs, 20)`) references `jobs` directly — correct.

**Test:** `NotificationCenter.test.tsx` precisely targets the regression symptom (job added post-mount must appear without reload) and uses `findByText` with Zustand `setState` for deterministic timing. Clean.

**E2E:** `.or()` + `await expect(reviewerControl).toBeVisible()` is the correct Playwright pattern — the old `Promise.race` on non-retrying `.isVisible()` could race to `null` and also missed the empty state added in #302.


---

PR #318, opened 2026-06-19
feat(blind-review): manager blind-review + per-kind reveal + shared compare view

raphaelfh:
## What

Managers now **review blind by default** and reveal other reviewers' answers **deliberately, per kind** (`extraction` / `quality_assessment`). One shared comparison view serves both screens, and the legacy frontend dual-read leak is deleted.

- **Per-kind setting** `projects.settings.managers_see_reviewers = {extraction, quality_assessment}` (both default `false`), written through a typed, manager-only endpoint `PUT /api/v1/projects/{id}/manager-review-visibility`.
- **Server-enforced, live** via `extraction_run_read_service.caller_can_see_peers(project_id, user_id, kind)`: consensus → always; manager → live per-kind setting; reviewer/viewer → never; finalized → all. Applied on both run read paths and the QA session open.
- **RLS 0025 unchanged** — a deliberate API-stricter-than-RLS split for the *manager* case only. The reviewer↔reviewer hard boundary stays lockstep at RLS + API + `resolve_caller_current_values`.
- **One shared `RunReviewerComparison`** for extraction + QA, fed by server-blinded `reviewerSummary.decisionsByCoord` (no direct Supabase read). The bespoke `loadValuesForOthers` dual-read and all legacy per-screen compare UI are removed (net ≈ −3000 lines there).
- **Per-kind toggle** mounted in both the extraction (Review/Consensus) and QA (Configuration) settings surfaces.

## Why

The post-`0025` blind-leak fix exempted managers (as arbitrators), which defeated blind review for a manager who also reviews. This makes manager blindness the default and the reveal an explicit, reversible, per-kind choice. See **ADR 0012** and the architecture doc §3/§5.

## Verification

- 532 frontend (vitest) + 14 blind-review backend integration tests (real local Postgres) + architectural fitness + typecheck + eslint — all green.
- Per-kind blind/reveal API E2E (`blind-review-manager.api.e2e.ts`): manager blind by default, live reveal, per-kind independence (extraction-on does not unblind QA), reviewer-always-blind.
- A final whole-branch review verified the three lockstep blind filters agree and caught a `projects.settings` dual-writer race that could revert the toggle — fixed here so the dedicated endpoint is the sole writer.

## Docs

- `docs/adr/0012-manager-blind-review-and-reveal.md`
- `docs/reference/extraction-hitl-architecture.md` (§3 RLS split, §5 reuse boundary, glossary)
- Design spec: `docs/superpowers/specs/2026-06-18-manager-blind-review-design.md`


raphaelfh:
## Claude review
**Verdict:** LGTM with nits
**Scope:** Introduces per-kind manager blind-review with server-enforced visibility toggle (`PUT /api/v1/projects/{id}/manager-review-visibility`), merges two per-screen comparison UIs into a single shared `RunReviewerComparison` component, and removes ~3 000 lines of legacy Supabase dual-read code.

> **Diff size note:** 84 files, ~6 400 lines — exceeds the 4 000-line review limit. Review focused on `backend/app/`, `frontend/{services,hooks,components}/`, `supabase/migrations/`, and `backend/alembic/` only. Generated docs and copy-key files not reviewed.

### Findings

- **Nit** `backend/app/services/manager_review_visibility_service.py` — Concurrent per-kind writes for the same project can silently race. Two in-flight requests (e.g. manager toggling `extraction` while a second request toggles `quality_assessment`) each read the same stale JSONB blob, modify one key independently, then both `flush()` — the later commit overwrites the earlier one. Fix: add `FOR UPDATE` to `ProjectRepository.get_by_id` when called from this service, or replace the read-modify-write with a single atomic `UPDATE projects SET settings = settings || jsonb_build_object('managers_see_reviewers', ...)`.

- **Nit** `frontend/components/extraction/manager-blind/ManagerReviewVisibilityToggle.tsx` — The toggle fires a plain `apiClient` PUT (no TanStack Query mutation), so no cache invalidation happens after a successful write. A manager who toggles and stays on the same run page will see stale server-blinded `reviewerSummary.decisionsByCoord` in the comparison view until they reload. Fix: call `queryClient.invalidateQueries({ queryKey: runsKeys.detail(runId) })` (or equivalent) in the `.then()` block of `setManagerReviewVisibility`.

### Incident-class sweep

| Class | Status |
|---|---|
| BOLA | ok — `require_project_manager` checks project membership + manager role in one atomic SQL predicate; no gap |
| TOCTOU | finding — settings JSONB concurrent-write race (nit above); run-state transitions untouched and use existing lifecycle service |
| Swallowed errors | ok — no silent-success catch; toggle component reverts optimistic state and toasts on error; `ProjectNotFoundError` correctly re-raised as 404 |
| ApiResponse envelope | ok — new endpoint returns `ApiResponse[ManagerReviewVisibilityRead]`; `apiClient` unwraps once; no double-unwrap |
| TanStack cache | finding — visibility toggle not cache-aware (nit above); no new cache keys missing scope variables |
| Schema drift | ok — `ManagerReviewVisibilityRead` booleans match JSONB structure; no Pydantic/nullable mismatch; missing keys resolved safely via `.get(kind, False)` |
| RLS | ok — RLS 0025 unchanged; API intentionally stricter than RLS for the manager case; reviewer↔reviewer hard boundary untouched |

---

Issue #42, opened 2026-05-04
bug(extraction_consensus): record_consensus select_existing publishes wrong value when selected_decision_id belongs to a different coordinate


raphaelfh:
## Summary

`ExtractionConsensusService.record_consensus` resolves the value to publish by fetching **all** reviewer decisions for the run (`list_by_run`) and then searching by `id` only. It never validates that `selected_decision_id.instance_id == instance_id` and `selected_decision_id.field_id == field_id`. A decision from a completely different `(instance_id, field_id)` coordinate pair in the same run can be selected, causing the wrong field's value to be published under the requested coordinate.

## File & Lines

`backend/app/services/extraction_consensus_service.py` lines 78–101

```python
decisions = await self._decisions.list_by_run(run_id)          # all decisions, all coords
selected = next((d for d in decisions if d.id == selected_decision_id), None)
# ↑ only filters by id — does NOT check d.instance_id or d.field_id
...
published_value = selected.value or {}
# accepted decision for (instance=A, field=X) now populates (instance=B, field=Y)
```

`_publish_internal` is then called with the **caller-supplied** `instance_id`/`field_id`, not the decision's own coordinates, so the mismatch is never detected downstream.

## Repro (3 lines)

1. In a CONSENSUS-stage run, reviewer submits two decisions: `dec_A` for `(instance=1, field=X)` and `dec_B` for `(instance=2, field=Y)`.
2. Call `POST /api/v1/runs/{id}/consensus` for `(instance=2, field=Y)` with `selected_decision_id=dec_A.id`.
3. `ExtractionPublishedState` for `(run, instance=2, field=Y)` is written with the value from `dec_A` (field X), not field Y's value. No error is raised.

## Verification

Code inspection:
- `ExtractionReviewerDecisionRepository.list_by_run` (`extraction_reviewer_decision_repository.py:23-30`) returns all rows for `run_id` with no coordinate filter.
- The `next(... if d.id == selected_decision_id ...)` predicate at line 79 only tests `d.id`.
- There is no DB-level composite FK or CHECK constraint on `ExtractionConsensusDecision` that would enforce coordinate matching (see also issue #20 for the cross-run variant).
- The existing `assert_coords_coherent` call at line 62 validates that `(run_id, instance_id, field_id)` is a coherent triplet, but does **not** validate the decision's own coordinates.

## Severity

[...]

PR #326, opened 2026-06-19
fix(tests): self-provision sibling coordinate in cross-coordinate consensus test

raphaelfh:
## Problem

`test_select_existing_rejects_cross_coordinate_decision` (integration) is under-constrained and behaves inconsistently:

- It searched for "a different `(instance, field)` coordinate in the same template" with a bare `LIMIT 1` and **no article scope**.
- The `#189` coherence guard (`assert_coords_coherent`, added *after* this `#42` test) requires the instance to belong to the **run's article**, not just the template. When the query returned a *different article's* instance, `record_consensus` raised `CoordinateMismatchError` **before** reaching the `"belongs to"` guard the test asserts → failure.
- Worse: in a **clean CI seed each article has exactly one field**, so the query returned nothing → `pytest.skip`. The test has effectively **never exercised the guard in CI**; it only "ran" locally when a shared/populated DB happened to contain a cross-article instance, at which point it failed.

This surfaced during a local preflight where the only red test was this one, against a dev DB shared across several worktrees.

## Fix

Stop borrowing a coordinate from the DB. Build a coherent sibling coordinate **deterministically**: add a field to the run instance's own `entity_type`, then record consensus on `(instance_id, other_field_id)` using the decision for `(instance_id, field_id)`.

- `(instance_id, other_field_id)` is coherent for the run (same article + template) → passes the coherence guard.
- It differs only in `field_id` → reliably trips the `"belongs to"` guard.
- No dependency on seed contents or DB state; the probe field rolls back with the test transaction (verified: 0 leaked rows).

## Verification

Run against the same populated local DB where the test previously failed:

- Target test ×3: `1 passed` each (no longer skips)
- `test_extraction_consensus_service.py` (integration): `10 passed`
- `test_extraction_consensus_service.py` (unit): `29 passed`
- `make test-backend`: `1857 passed, 3 skipped, 0 failed`
- `ruff check` + `ruff format --check`: clean

### DIFF
--- a/backend/tests/integration/test_extraction_consensus_service.py
+++ b/backend/tests/integration/test_extraction_consensus_service.py
@@ -341,34 +341,42 @@ async def test_select_existing_rejects_cross_coordinate_decision(
         pytest.skip("Missing fixtures.")
     run_id, instance_id, field_id, profile_id, decision_id = fx
 
-    # Find a different (instance, field) coordinate in the same template.
-    other_row = await db_session.execute(
-        text(
-            """
-            SELECT i.id, f.id
-            FROM public.extraction_instances i
-            JOIN public.extraction_entity_types et ON et.id = i.entity_type_id
-            JOIN public.extraction_fields f ON f.entity_type_id = et.id
-            WHERE (i.id <> :iid OR f.id <> :fid)
-              AND i.template_id = (
-                  SELECT template_id FROM public.extraction_instances WHERE id = :iid
-              )
-            LIMIT 1
-            """
-        ),
-        {"iid": instance_id, "fid": field_id},
-    )
-    other = other_row.first()
-    if other is None:
-        pytest.skip("Need a second coordinate in the same template.")
-    other_instance_id, other_field_id = other
+    # Build a second coordinate that is coherent for this run by adding a field
+    # to the run instance's own entity_type. (instance_id, other_field_id) then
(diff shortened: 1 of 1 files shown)


raphaelfh:
## Claude review
**Verdict:** LGTM
**Scope:** Fixes `test_select_existing_rejects_cross_coordinate_decision` — previously skipped in CI (clean seed has exactly one field per article, so the cross-template `LIMIT 1` query returned nothing) and flaky locally (returned a different article's instance, causing `CoordinateMismatchError` before the "belongs to" guard under test). Fix self-provisions a sibling field on the run instance's own entity type so the probe coordinate is coherent for the run and reliably trips the right guard.

### Findings

No findings above the confidence bar — the deterministic CI gates cover the rest.

### Incident-class sweep

| Class | Status |
|---|---|
| BOLA | n/a — test-only change, no endpoint touched |
| TOCTOU | n/a |
| Swallowed errors | ok — `.scalar()` returns `None` on failure but upstream `fx` guard already validates `instance_id`; `INSERT … RETURNING id` always returns a value |
| TanStack cache | n/a |
| Schema drift | n/a — `'text'` is a valid `extraction_field_type` enum variant (confirmed in `baseline_v1.sql:37`); no UNIQUE constraint on `(entity_type_id, name)` in `ExtractionField`, so parallel test runs won't conflict |
| RLS | n/a |

---

PR #328, opened 2026-06-19
fix(e2e): unblock QA suite in CI + de-flake export/consensus/blind-review

raphaelfh:
## What

Test-infrastructure only (no product/backend code). Makes the QA / multi-reviewer-consensus / blind-review e2e suite actually run in CI and de-flakes it.

## Why

While verifying all flows post-deploy, the local e2e suite failed 3 tests that are **green in CI** — because CI silently **skips** the whole QA suite: `E2E_QA_GLOBAL_TEMPLATE_ID` had no zero-config default, so the consensus + blind-review contracts only ran on warm local environments, where they then flaked or asserted stale behavior. Root causes, each fixed at the right layer:

1. **CI silent-skip coverage hole** — default `E2E_QA_GLOBAL_TEMPLATE_ID` to the seeded PROBAST global id in `playwright.config.ts`, mirroring the other zero-config fixture defaults. The QA/consensus/blind-review tests now run in CI.
2. **Stale 3-way consensus expectation** — under blind-review (#318) a manager is blinded to peer reviewer decisions in the `consensus` stage by design; the backend returning only the caller row is **correct**. The test now performs the manager **reveal** (`PUT /manager-review-visibility`, QA=true) before reading the divergence, matching the real arbitration flow and `blind-review-manager.api`.
3. **Cross-file fixture race** — the two suites that hard-reset QA runs (`resetQaRuns` deletes every QA run for a (project, article) pair) shared one article, so in parallel one suite deleted the other`\s in-flight run mid-decision. Each now uses its own dedicated fixture article (neutral names to avoid colliding with the export dialog`\s `getByLabel` locators).
4. **prepareCleanQaRun** advance-to-review made idempotent (tolerate a concurrent `review->review`).
5. **export UI-flow** made serial so parallel same-account logins don`\t trip the local Supabase auth rate limit.

## Verification

- Full local `test:e2e:local`: **54 passed / 0 failed** (was 49 / 3 failed).
- consensus + blind-review **3x stress** at `--retries=0`, full parallelism: green (previously 2/3 failed).
- eslint on changed files: clean.

### DIFF
--- a/frontend/e2e/_fixtures/ensure-fixtures.ts
+++ b/frontend/e2e/_fixtures/ensure-fixtures.ts
@@ -142,6 +142,16 @@ export async function ensureFixtures(): Promise<void> {
   await ensureMembership(F.PROJECT_ID, reviewerCId, "reviewer");
   await ensureArticle(F.ARTICLE_ID, F.PROJECT_ID, "E2E Fixture Article");
   await ensureArticleText(F.PROJECT_ID, F.ARTICLE_ID);
+
+  // Dedicated articles for the QA API suites that hard-reset runs, so their
+  // parallel `resetQaRuns` calls can never delete each other's in-flight run.
+  // Names are intentionally neutral: these rows show up in the project's
+  // extraction list, and a substring like "Consensus"/"Single user" would
+  // collide with the export dialog's getByLabel() locators.
+  await ensureArticle(F.QA_CONSENSUS_ARTICLE_ID, F.PROJECT_ID, "E2E QA Article One");
+  await ensureArticleText(F.PROJECT_ID, F.QA_CONSENSUS_ARTICLE_ID);
+  await ensureArticle(F.QA_BLIND_REVIEW_ARTICLE_ID, F.PROJECT_ID, "E2E QA Article Two");
+  await ensureArticleText(F.PROJECT_ID, F.QA_BLIND_REVIEW_ARTICLE_ID);
   const ownerToken = await login(F.OWNER_EMAIL, F.FIXTURE_PASSWORD);
   await ensureCharmsImported(F.PROJECT_ID, ownerToken);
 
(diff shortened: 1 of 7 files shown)


raphaelfh:
## Claude review
**Verdict:** LGTM with nits
**Scope:** Test-infrastructure only — de-flakes the e2e QA suite by adding a zero-config `E2E_QA_GLOBAL_TEMPLATE_ID` default in `playwright.config.ts` (unblocking the whole QA suite in CI), giving each hard-resetting suite its own dedicated fixture article to prevent cross-suite `resetQaRuns` collisions, making `prepareCleanQaRun` stage-advance idempotent under concurrent use, inserting a manager reveal before asserting consensus divergence (matching the real arbitration flow), and serialising the export UI tests to avoid local Supabase auth rate-limit flakes.

### Findings

- **nit** `frontend/e2e/flows/qa-multi-reviewer-consensus.api.e2e.ts:249–266` — `setQaVisibility(true)` is not protected by `try/finally`. If any assertion between the reveal and the restore fails (e.g. `expect(reviewers.size).toBe(3)` at line 262), the project is left with `managers_see_reviewers=true` for subsequent retries or CI re-runs of the suite. Because sibling blind-review cases test the blinded default, leaked state can mask failures there. This is a plain Playwright test helper, not a React component, so the React Compiler `try/finally` restriction does not apply. Suggested fix:

  ```typescript
  await setQaVisibility(true);
  try {
    // existing assertions (detailRes → expect(reviewers.size).toBe(3))
  } finally {
    await setQaVisibility(false);
  }
  ```

### Incident-class sweep

| Class | Status |
|---|---|
| BOLA | n/a — no product endpoints added or modified |
| TOCTOU | ok — `prepareCleanQaRun` race addressed correctly: advance only fails when end state is not `review` |
| Swallowed errors | ok — conditional suppression in `hitl.ts:159–170` is correctly scoped to "desired state already reached" |
| ApiResponse envelope | n/a — no product API calls changed; existing `parseEnvelope` usage untouched |
| TanStack cache | n/a |
| Schema drift | n/a |
| RLS | n/a |

---

PR #330, opened 2026-06-20
Promote dev → main: fix extraction "No fields" masking (#329)

raphaelfh:
Promotes `dev` → `main` (production: Railway backend + Vercel frontend).

## Contents (dev ahead of main by 2 commits)
- **#329** `fix(extraction): surface run-open failures instead of masking them as "No fields"` — the extraction view rendered "No fields for extraction" whenever the RunView/session-open failed or was in flight (a #324 regression), masking the real failure. Now the gate reports `no-fields` only when the run is loaded and genuinely empty; a missing run is a loader or a retryable error.
- **#328** `fix(e2e): unblock QA suite in CI + de-flake export/consensus/blind-review` (already on dev).

## Verification
- All 8 required checks green on dev HEAD (`31b8b3b5`): Backend Lint/Tests/E2E/Docker, Frontend Lint/Build/Tests, Architectural Fitness.
- #329: 647 frontend unit tests + 10 new gate-scenario tests; typecheck/eslint/fitness clean.
- Browser-verified against the prod backend: the masked "No fields" condition now renders "Could not open this extraction" + Retry.
- Note: `Frontend E2E (ephemeral stack)` is non-required and showed a **pre-existing flaky** failure in `qa-reopen.ui.e2e.ts` (QA reopen flow — a different page with no shared code with #329; the prior commit #328 was itself a QA-suite de-flake).


raphaelfh:
## Claude review
**Verdict:** LGTM with nits
**Scope:** Promotes dev → main with two fixes: surfaces extraction run-open failures as a retryable error state (replacing the false "No fields" empty state from regression #324) and de-flakes the QA/export E2E suites by giving each resetting suite its own fixture article and guarding the concurrent `advance` call.

### Findings

- **minor** `frontend/e2e/flows/qa-multi-reviewer-consensus.api.e2e.ts:258–261` — `setQaVisibility(false)` is called inline in the test body, not in a `try/finally`. If any assertion between `setQaVisibility(true)` and `setQaVisibility(false)` throws (e.g. `expect(reviewers.size).toBe(3)`), the project-level `managers_see_reviewers[quality_assessment]` setting stays `true` for the rest of the serial run, potentially causing downstream cases in the same suite to see reviewer decisions they expect to be blinded. Suggested fix: wrap the reveal-assert-restore block in `try/finally`:
  ```ts
  await setQaVisibility(true);
  try {
    // ... detailRes fetch + assertions ...
  } finally {
    await setQaVisibility(false);
  }
  ```

### Incident-class sweep

| Class | Status |
|---|---|
| BOLA | n/a — pure frontend + E2E; no new endpoints |
| TOCTOU | ok — the `stageOf()` re-check after a failed advance is the right pattern for a fixture helper |
| Swallowed errors | ok — advance failure is swallowed only when the run is already at the target stage; real failures still propagate |
| ApiResponse envelope | ok — `parseEnvelope` usage unchanged |
| TanStack cache | ok — retry calls `refetch()` on both hooks; cache keys unchanged |
| Schema drift | n/a — no model/migration/type changes |
| RLS | n/a — no migrations |

---

PR #332, opened 2026-06-20
Promote dev → main: AI-accept spinner fix (#331)

raphaelfh:
Promote `dev` → `main`: ships the stuck AI-accept spinner fix (#331).

- `ExtractionFormView` memo comparator now tracks `isActionLoading`, so the post-accept `clearLoading` re-render reaches `FieldInput` and the spinner clears instead of spinning forever.
- Symmetric cleanup of the accept/reject handlers; consolidated end-to-end spinner regressions (extraction accept/reject, model-child sub-tree, QA path) + comparator unit test, adversarially validated.

All `dev` CI checks were green on #331. Frontend-only change.

### DIFF
--- a/frontend/components/extraction/ExtractionFormView.tsx
+++ b/frontend/components/extraction/ExtractionFormView.tsx
@@ -157,6 +157,10 @@ export const ExtractionFormView = memo(ExtractionFormViewComponent, (prevProps, 
     prevProps.modelChildSections === nextProps.modelChildSections &&
     prevProps.activeModelId === nextProps.activeModelId &&
     prevProps.models.length === nextProps.models.length &&
+    // `isActionLoading` gets a new ref on every `actionLoading` change. Omitting
+    // it lets the memo swallow the post-accept `clearLoading` (its only changed
+    // prop), so the spinner never clears. See FieldInput's comparator.
+    prevProps.isActionLoading === nextProps.isActionLoading &&
     !aiSuggestionsChanged
   );
 });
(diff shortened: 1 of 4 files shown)


raphaelfh:
## Claude review
**Verdict:** LGTM with nits
**Scope:** Adds `isActionLoading` to the `ExtractionFormView` memo comparator so the post-accept `clearLoading` re-render propagates to `FieldInput` and clears the spinner; cleans up `console.warn` comments in accept/reject handlers; ships a unit test for the comparator and a 3-path E2E regression suite.

### Findings

- **nit** `frontend/hooks/extraction/ai/useAISuggestions.ts:332` — `isActionLoading` is a plain inline function with no `useCallback`. The comparator's `prevProps.isActionLoading === nextProps.isActionLoading` check is only meaningful if the React Compiler memoizes it against `actionLoading` state — which it should under the project's `panicThreshold: 'all_errors'` config. But the invariant is implicit: if this hook ever acquires a `'use no memo'` directive or a compiler bailout, the comparator would always be `false`, silently defeating the memo guard (spinner still clears; perf loss only). An explicit `useCallback([actionLoading])` would make the contract visible without requiring compiler introspection.

- **nit** `frontend/test/spinner-fix.e2e.test.tsx:1–27` — 22-line multi-paragraph file header; project rule is "never write multi-paragraph docstrings or multi-line comment blocks." Not a blocker.

### Incident-class sweep

| Class | Status |
|---|---|
| BOLA | n/a — frontend-only, no new endpoints |
| TOCTOU | n/a — no backend state transitions |
| Swallowed errors | ok — no new error handling paths |
| ApiResponse envelope | n/a — no API calls changed |
| TanStack cache | n/a — no queries or mutations touched |
| RLS | n/a — no migrations |

---

PR #345, opened 2026-06-20
ci(docs): define + enforce the per-layer status enum (WS-3)

raphaelfh:
## What

**WS-3 of the governance cleanup** — turns the doc status vocabulary from a
narrated, drifted convention into a **CI-enforced, per-layer enum**. Off `dev`
(which already has WS-4), so isolated.

The audit found `docs/README.md` declared 6 status values but the corpus used
**14** across ADRs/specs/plans/reference, and `check-frontmatter.sh` only checked
key *presence*, never the value (A5/A5b) — plus one bare `owner` (G3).

## Changes

- **`docs/README.md`** — replaces the flat enum with **per-layer subsets**:
  - Reference / how-to: `stable · draft · deprecated`
  - ADRs (MADR): `proposed · accepted · rejected · deprecated · superseded · template`
  - Specs / plans: `draft · approved · in_progress · shipped · superseded · frozen`
- **`scripts/docs/check-frontmatter.sh`** — now rejects a status value outside its layer's enum, or an `owner` that isn't an `@handle` (archived docs stay exempt). "Enforced, not narrated."
- **Normalized the remaining active frontmatter** to conform:
  `planned→draft` (the 2 parsing plans), `ready→approved` (runheader ×2 + extraction-view-ux-p0), `proposed→draft` (ux-design spec), `stable→frozen` (autoloop-design — a settled living design; `frozen` keeps it on the active surface as an immutable reference, like the HITL design spec), and the lone bare owner `raphaelfh → '@raphaelfh'`.

## Verification

- New gate **passes** on the whole tree.
- **Negative test**: injecting a bogus status + a bare owner is caught (`BAD status …` / `BAD owner …`) — proving the gate actually enforces, not just passes.
- `markdownlint` — 0 · `cspell` — 0.

## Findings resolved

A5, A5b, G3.

## Note

Once this lands, an out-of-enum status or a bare owner **fails CI** — so the
14-values-across-the-corpus drift can't recur. Two docs (`autoloop-design`,
`observability-extraction`) are living-reference content sitting in spec/how-to
dirs; relocating them into the right Diátaxis quadrant is a WS-5 follow-up
(here they're just made enum-conformant in place).


raphaelfh:
## Claude review
**Verdict:** LGTM
**Scope:** Turns the docs status vocabulary from a narrated convention into a CI-enforced per-layer enum by upgrading `check-frontmatter.sh` to validate status values and owner format, and normalising 8 existing docs to conform.

### Findings

### Incident-class sweep

| Class | Result |
|---|---|
| BOLA | n/a — no API endpoints changed |
| TOCTOU | n/a — no run-state logic |
| Swallowed errors | ok — failures correctly set `FAIL=1`; `[[ -n "$status_val" ]]` guard avoids false-positives on missing keys without hiding real violations |
| TanStack cache | n/a |
| Schema drift | n/a — no SQLAlchemy / Pydantic / TS types touched |
| RLS | n/a — no migrations |

**Shell-script logic spot-check:**
- `awk` frontmatter extraction correctly scopes to between first and second `---` delimiters. ✓
- `owner_re=^'?@[A-Za-z0-9_-]+'?$` handles both unquoted `@handle` and single-quoted `'@handle'` (the in-repo convention). ✓
- `case` statement correctly covers ADR / specs+plans / everything-else layers; bash `|` in a case item is valid. ✓
- All 8 frontmatter normalizations map to their layer's allowed set (`planned→draft`, `ready→approved`, `proposed→draft`, `stable→frozen`). ✓


---

PR #343, opened 2026-06-20
chore(governance): WS-4 archive sweep — graduate 29 shipped docs out of the read-path

raphaelfh:
## What

**WS-4 of the governance cleanup** — the archive sweep. Per the ratified SOTA
decision (Q1), shipped plans/specs **graduate out of the agent read-path** so
agents stop loading completed work as context (Git keeps the full history). Off
`dev`, so this diff is isolated from [#342](https://github.com/raphaelfh/prumo/pull/342) (WS-1/WS-2).

## Changes

- **Archived 29 shipped docs** → `docs/superpowers/{plans,specs}/archive/2026-06-20-governance-sweep/` (status set to `shipped`):
  14 verified-shipped plans (runopen-slowload ×2, react-compiler ×2, llm-stack, manager-blind-review, blind-review-cleanup, in-app-feedback, e2e-fixture, xlsx-export, run-vocabulary, data-path-finish, review-stage, pollution-cleanup, linear-integration) + their 14 sibling design specs + the shipped UI specs (sidebar, pdf-viewer-db, dark-light, preflight).
- **Kept active** (not archived): parsing / grounded-extraction, dev-workflow-sota, the runheader + extraction-view-ux work (merged via #338), the **frozen** HITL design spec (its graduation is WS-6), and the `stable`/uncertain specs.
- **Repointed 6 inbound provenance paths** (ADRs 0010/0011/0012, `extraction-hitl-architecture.md`, `test-infra-phase3-kickoff.md`) to the archive location — zero stale active-surface pointers remain.
- **CLAUDE.md "Current focus" + ROADMAP** no longer name the shipped data-path consolidation as active: focus collapses to a ROADMAP pointer at the genuinely-active parsing work (≤5-line budget honored); consolidation moved to "Recently shipped"; drifted visible dates fixed.
- **`.markdownlintignore` pruned** to active plans only (archived files are glob-covered); also removed the 7 stale G6 entries — every non-glob entry now resolves to a tracked file.

## Findings resolved

B1, B2, B3, B4, B5, B6, B7, B8, B9, B10, B11, B12, B13, E2, F5 (via archival), D4 (partial), **G6**.

## Verification

- `markdownlint-cli` (the docs-ci gate) — **exit 0, 0 violations**
- `cspell` — **0 issues**
- `check-frontmatter.sh` — passed (archived files exempt; active files valid)
- `0` stale old-path references on the active surface after the moves
- every non-glob `.markdownlintignore` entry resolves to a tracked file

## Notes

[...]

PR #346, opened 2026-06-20
docs(plans): dual-path parse-at-ingest + markdown-view/markdown-citations

raphaelfh:
## What

Two sequential, docs-only implementation plans for the extraction overhaul (authored via the `writing-plans` skill). No code changes.

- **`2026-06-20-dual-path-parse-at-ingest.md`** (7 phases, 12 tasks) — at ingest, every article (any route — Zotero, future direct-upload) is parsed into the existing `article_text_blocks` by a pluggable `DocumentParser` chosen per project: **Docling** (standard, self-hosted) or **LlamaParse** (premium, cloud, BYOK, activated per-project in settings), behind `create_document_parser()` with a fail-closed PHI gate. Adds the Celery `parse_article_file_task` + a single `ArticleFileIngestService` hook covering all routes. Stops at "blocks populated".
- **`2026-06-20-markdown-view-and-markdown-citations.md`** (7 phases, 16 tasks) — `render_blocks_to_markdown(blocks) → (markdown, source_map)` (reuses the assembler's `_render_table`); the AI reads the markdown; citations are anchored **in the markdown** (primary, robust, parser-independent) and projected onto the PDF bbox via the source-map (secondary, precision tracking the parser tier). Repoints the viewer's reader mode to a real markdown view, retires the 15k `pypdf` truncation, and **refines ADR-0013** (markdown gains its own anchor space). Depends on plan 1.

## Decisions baked in (confirmed with the owner)

- Citation surface = **markdown-primary + PDF projection** (refines ADR-0013's canvas-only highlight limitation).
- Standard self-hosted parser = **Docling**; premium = **LlamaParse**; pluggable via `PARSER_BACKEND` (the Phase-0 bake-off can re-rank).
- Two sequential plans; each ends with a **no-legacy cleanup + `/simplify`** pass.

## Notes

- Both plans carry the required frontmatter (`status`/`last_reviewed`/`owner`), the subagent-driven-development header, Global Constraints, a File map, TDD checkbox steps with real interface names, and a Self-Review. Registered in `.markdownlintignore`; new code-side terms added to `.github/cspell-words.txt`.
- Anthropic-provider config was deliberately kept OUT of plan 1 (it belongs to the separate vision / second-provider plan — adding it now would be unused config).


raphaelfh:
## Claude review
**Verdict:** LGTM with nits
**Scope:** Adds two sequential implementation plans for pluggable dual-path PDF parsing at ingest (Docling/LlamaParse + PHI gate) and a follow-up markdown projection + citation system; also cleans up `.markdownlintignore` and registers new spell-check terms.

### Findings

- **Minor** `docs/superpowers/plans/2026-06-20-dual-path-parse-at-ingest.md` Task 1.2/1.3 — The plan uses the phrase "fail-closed PHI gate" but the implementation defaults `Project.is_phi = False` (non-PHI), so new projects can receive LlamaParse without any explicit opt-out. The gate IS fail-closed in the sense that PHI-marked projects always stay self-hosted and missing keys always fall back to Docling, but the terminology "fail-closed" typically implies that the safe default is the most-restrictive state. An implementer reading the prose and then seeing `server_default=sa.false()` may be confused. Consider renaming to "explicit-PHI gate" or adding a one-liner comment: `# new projects are non-PHI by default; set is_phi=True to force self-hosted` — wording is cheap to fix at the plan stage.

- **Minor** `.markdownlintignore` line 38 — The cleanup removes the explicit ignore entry for `docs/superpowers/specs/2026-04-28-pdf-viewer-database-requirements.md`. If this spec was not graduated to `docs/superpowers/specs/archive/` in the WS-4 sweep (#343), it will now be linted and may fail. CI would catch it before merge, but worth verifying the spec was archived.

### Incident-class sweep

| Class | Status |
|---|---|
| BOLA | ok — `require_project_manager` on `PUT .../parser-settings`; `ensure_project_member` + 403 test on `GET .../markdown`; BOLA cited explicitly in plan 2 global constraints |
| TOCTOU | n/a — no run-state machine changes; Celery task uses `worker_session()` + `run_task()` + `self.retry` per the project pattern |
| Swallowed errors | ok — parser failures raise `ValueError`; Celery retries with `self.retry(exc=exc)`; no catch-and-return-success |
| ApiResponse envelope | ok — `ApiResponse.success(...)` at every response site; API-contract regen step included in Tasks 4.2 of both plans |
| TanStack cache | n/a — no frontend cache changes in this PR; plan 2 adds `articleKeys.markdown(id)` with a properly scoped key |
| Schema drift | ok — `MarkdownCitationAnchor` and `MarkdownRange` are mirrored verbatim in Python (`extraction.py`) and TypeScript (`citation.ts`); API-types regen is a required step |
| RLS | ok — `is_phi` column lands on `projects` table (already RLS-protected); no new table added; plan 2 explicitly notes "NO migration" |

---

PR #356, opened 2026-06-21
Promote dev → main: parse-to-markdown pipeline activation (PR-B)

raphaelfh:
## Promote dev → main

Ships **PR-B — parse-to-markdown pipeline activation** ([#355](https://github.com/raphaelfh/prumo/pull/355)) to production. This is the only change on `dev` not yet on `main`.

**Production effect (Fase 0):** the Railway worker that consumes `article_files.extraction_status = pending` goes live, so newly-uploaded article PDFs are parsed to `article_text_blocks`. Manual "Add article" / file uploads now enqueue the parse (previously they bypassed the backend), and a per-file **Re-parse** action recovers stuck/failed files. Terminal `parse_failed` is now durable.

No schema migration. The browser→Supabase Storage upload path is unchanged; only the `article_files` row creation moved server-side (BOLA-gated + storage-key validated).

### DIFF
--- a/.markdownlintignore
+++ b/.markdownlintignore
@@ -16,6 +16,7 @@ docs/superpowers/plans/2026-06-20-dual-path-parse-at-ingest.md
 docs/superpowers/plans/2026-06-20-markdown-view-and-markdown-citations.md
 docs/superpowers/plans/2026-06-20-extraction-header-refinement.md
 docs/superpowers/plans/2026-06-20-table-view-ux-consolidation.md
+docs/superpowers/plans/2026-06-20-parse-pipeline-activation-prb.md
 
 # Design specs — likewise snapshots
 docs/superpowers/specs/2026-*-design.md
--- a/backend/app/api/v1/endpoints/article_files.py
+++ b/backend/app/api/v1/endpoints/article_files.py
@@ -1,0 +1,104 @@
+"""Article-file ingest + recovery endpoints.
+
+The browser uploads the bytes to Supabase Storage under its own JWT; these
+endpoints create/recover the DB row server-side (via ArticleFileService) so
+membership is enforced and the parse is scheduled — the direct PostgREST
+insert that bypassed parsing is retired.
+"""
+
+from __future__ import annotations
+
+from uuid import UUID
+
+from fastapi import APIRouter, Depends, HTTPException, Request, status
+
+from app.api.deps.security import ensure_project_member, get_current_user_sub
+from app.core.deps import DbSession
+from app.schemas.article import ArticleFileResponse, ConfirmUploadRequest
+from app.schemas.common import ApiResponse
+from app.services.article_file_service import ArticleFileService, ParseEnqueueError
+from app.services.article_text_block_read_service import (
+    ArticleFileNotFoundError,
+    get_article_file_project_id,
+)
+from app.services.citation_read_service import ArticleNotFoundError, get_article_project_id
+
+router = APIRouter(tags=["article-files"])
+
+_ENQUEUE_FAILED_DETAIL = "Failed to schedule parsing; please retry"
+
+
+def _trace(request: Request) -> str | None:
+    return getattr(request.state, "trace_id", None)
+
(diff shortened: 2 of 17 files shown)


raphaelfh:
## Claude review
**Verdict:** Needs attention
**Scope:** Activates the parse-to-markdown pipeline by adding two server-side endpoints (`POST /articles/{article_id}/files` to confirm browser→Storage uploads and `POST /article-files/{article_file_id}/reparse` to re-queue failed files), removing the direct `supabase.from('article_files').insert` from the frontend, and wiring terminal Celery retry failures to a durable `_mark_parse_failed` helper. No schema migration; no RLS changes.

### Findings

- **important** `backend/app/api/v1/endpoints/article_files.py:~73` — Storage-key prefix check uses `startswith(f"{project_id}/{article_id}/")`, which does **not** reject keys containing `..` path segments. A project-A member could upload to `{projA_id}/{artA_id}/../{projB_id}/{artB_id}/evil.pdf` and the check would pass. Whether this is exploitable depends on whether Supabase Storage normalises object-key paths (S3-compatible stores typically don't, but that's undocumented here). Suggested fix: add a one-liner before the prefix check — `if any(seg in (".", "..") for seg in body.storage_key.split("/")):  raise HTTPException(400, "storage_key must not contain path traversal segments")`.

- **important** `backend/app/services/article_file_service.py:~239` — `ArticleFileService.reparse` unconditionally resets `extraction_status` to `"pending"` for any file regardless of current state. The PR description says the Re-parse action "recovers stuck/failed files", but the endpoint allows re-parsing a file already in `parsed` state. If the upstream Celery task clears `article_text_blocks` before re-inserting (unknown from this diff), triggering re-parse on an active-`parsed` file could delete blocks that an in-flight extraction run is consuming, silently corrupting it. Suggested fix: guard with `if article_file.extraction_status not in ("parse_failed", "pending"): raise HTTPException(400, "only failed or stuck files may be re-parsed")`, or add a test that explicitly asserts re-parsing a `parsed` file is safe (proving the task upserts, not clear-then-insert).

- **nit** `frontend/components/articles/ArticleDetailDialog.tsx` — The Re-parse button calls `reparseArticleFile` and shows a toast, but there is no `queryClient.invalidateQueries` after success. The button stays active and `extraction_status` stays stale in the UI, so a user could click Re-parse multiple times in rapid succession and enqueue duplicate parse tasks. Suggest invalidating the article-files cache key (even if polling is deferred to PR-D) to at least reflect the `pending` state immediately after the call.

- **nit** `backend/app/worker/tasks/parsing_tasks.py:~290` — `_mark_parse_failed` contains module-level imports (`from sqlalchemy import select`, `from app.models.article import ArticleFile`, `from app.worker._session import worker_session`) placed inside the function body. This is unusual, hides circular-import issues at startup, and isn't consistent with the rest of the file. Move them to the top of the module.

### Incident-class sweep

| Class | Status |
|---|---|
| BOLA | ok — both endpoints call `ensure_project_member`; `project_id` is resolved server-side from `article_id` / `article_file_id`, never from request body; storage-key prefix ties the object to the resolved project (with the `..`-segment caveat above) |
| TOCTOU | ok — no run-state machine touched; `reparse` status reset is a simple assignment on a single row with negligible race window |
| Error swallowing | ok — `_enqueue_or_fail` catches broadly but persists failure durably and re-raises as `ParseEnqueueError`; Celery terminal-retry path correctly captures `error_message` before exception scope exits |
| Envelope | ok — both endpoints declare `ApiResponse[ArticleFileResponse]`; frontend service uses `apiClient` (not bare Supabase client) for both new calls |
| TanStack cache | finding — Re-parse mutation lacks `invalidateQueries`; button stays live and stale |
| Schema drift | nit — `extraction_status` as free-text string literals in three places; MF-11 explicitly deferred to PR-C, so noting but not blocking |
| RLS | ok — no new tables; service-role DB connection with explicit `ensure_project_member` gate is the established pattern; no new RLS bypasses |

---

PR #369, opened 2026-06-21
feat(extraction): collapse run-stage proposal+review → extract (HITL Phase 1)

raphaelfh:
## HITL Phase 1 — collapse the extraction `proposal` + `review` stages into `extract`

Finishes the cross-cutting rename begun by the enum migration (`0028`) +
lifecycle work on this branch. The run lifecycle becomes
`pending → extract → consensus → finalized` (+ `cancelled`). Supersedes
**ADR-0010** (new **ADR-0014**).

### Why
`review` modelled a *user activity* as a lifecycle stage, reached by a
client-side auto-advance (`useAutoAdvanceToReview`, with a documented race +
"advance failed" toast) and bridged by an invisible `proposal → review`
boundary materialization. Collapsing to a single editable `extract` stage —
where humans write per-user `ReviewerDecision`s directly — deletes all of that
machinery.

### Backend
- `record_decision` gates on `extract`; the proposal gate accepts AI/system
  proposals in `extract` and **rejects human `extraction` writes** (blind-review
  write defense — they must use `/decisions`). `model_hierarchy` `modelling_method`
  now writes a `ReviewerDecision` instead of a shared human proposal.
- Deleted `_materialize_human_decisions` (no `proposal → review` edge to bridge).
- Collapsed every `create → proposal` advance to `create → extract`; dropped the
  dead `proposal → review` advances. `auto_advance_to_review` kept for API compat
  but inert.
- Stage lists/gates, `AdvanceStageRequest` regex, the all-users export omit
  filter, and `POSTGRESQL_ENUM_VALUES` all move to `extract`.

### Frontend
- `ExtractionRunStage` union → `pending|extract|consensus|finalized|cancelled`.
- The editable-stage hooks route by run **kind**: in `extract`, extraction
  autosaves to `/decisions` (per-user) and reads back from `currentValues`; QA
  stays on `/proposals` + raw-proposals hydration (`kind` threaded into
  `useAutoSaveProposals` + `useExtractedValues`).
- Deleted `useAutoAdvanceToReview` (+ test) and its wiring; the run sits in
  `extract` until "Mark ready" advances to consensus.

### Docs / types
- **ADR-0014** + `extraction-hitl-architecture.md` (lifecycle diagrams, §2.2
  stage table, "Stage advance", enum table). Regenerated `frontend/types/api`;
  applied only the `extraction_run_stage` enum delta to the Supabase types by
  hand (a full local regen would pull in unrelated sibling-worktree schema from
  the shared dev DB).

### Verification
[...]

PR #372, opened 2026-06-21
fix: post-merge review — criticals, BOLAs, edit-loss, QA finalize + cleanups

raphaelfh:
## Why

Post-merge adversarial review of the batch on `dev`, all findings verified by independent skeptic agents (0 refuted). **Rebased onto current `dev`** (post HITL Phase 1–3 / #369/#374/#375) and re-triaged: every fix below was re-confirmed still present on the latest `dev`; two original findings turned out STALE (now-live code) and were deliberately left alone.

## What changed

**🔴 Critical**
- `fix(export)` — role-first `_build_tidy_tables`: cardinality=ONE `MODEL_SECTION`s rendered zero rows (silent data loss in every prediction-model export).
- `fix(pdf)` — `usePageHandle` degrades outside a `ViewerProvider` instead of throwing → QA "jump to source" / AI suggestion modal no longer crashes to blank.

**🟠 Security — two confused-deputy BOLAs**
- `resolve_form_runs` and `_load_active_template_version` now scope by `project_id` (form-runs run-id leak; async-export template-metadata leak).

**🟠 Important**
- `fix(extraction)` — `saveNow()` now rejects on autosave failure so the `onMarkReady`/`onOpenConsensus` guards block a run-advance past unsaved edits.

**🟡 Minor**
- `fix(qa)` — QA passes `isComplete` to `ConsensusPanel` so a no-divergence run can finalize (was wrongly disabled + "incomplete").
- `chore(cleanup)` — remove dead `getRoleLabel`; English fallback + drop PT/emoji debug log in `useModelManagement`; `CitationOverlay.focus({preventScroll})`; null-guards in `aiSuggestionService`.

## Test plan (all verified locally; local DB synced to `dev` head `0030`)

- [x] export tidy-table: 2 unit tests + full export suite (275) — TDD red→green.
- [x] PDF crash: no-provider vitest (real hook chain) — TDD red→green; pdf-viewer suites (32).
- [x] BOLA #3/#4: integration regression tests — TDD red-checked (fail when the `project_id` predicate is removed).
- [x] autosave: regression test mirroring the caller guard + 2 updated error-path tests — TDD red→green (23).
- [x] QA + cleanups: typecheck + eslint clean; affected suites (114) green; no regressions.

## Deliberately NOT changed (re-triaged)

- `PARSER_BACKEND`, `is_run_arbitrator` — review flagged them dead, but they are LIVE again on current `dev` (#359/#371/#374).
- `study_instances` — a documented read-compat alias load-bearing for ~15 test call sites, not dead.

## Out of scope (behavior-changing minors, deferred for a decision)

- LLM schema `number`/`integer`→`float` coercion; hard `Literal` allowed-values constraint.
- `validateFieldImpact` append-only overcount; `reparse` in-flight guard.

### DIFF
--- a/backend/app/api/v1/endpoints/articles.py
+++ b/backend/app/api/v1/endpoints/articles.py
@@ -106,7 +106,9 @@ async def post_form_runs(
     """
     await ensure_project_member(db, body.project_id, current_user_sub)
 
-    refs = await resolve_form_runs(db, body.article_ids, template_id=body.template_id)
+    refs = await resolve_form_runs(
+        db, body.article_ids, project_id=body.project_id, template_id=body.template_id
+    )
     return ApiResponse.success(refs, trace_id=_trace(request))
 
 
--- a/backend/app/services/extraction_export_service.py
+++ b/backend/app/services/extraction_export_service.py
@@ -1948,14 +1948,24 @@ def _build_tidy_tables(
     value_map: dict[tuple[Any, ...], Any],
     mode: ExportMode,
 ) -> tuple[TidyTable, ...]:
-    """One publication table per non-container section at its cardinality grain.
+    """One publication table per non-container section at its record grain.
 
-    ``cardinality==MANY`` fans out one row per (article × instance) for ANY
-    role (spec §5.2 — never a ``role==MODEL_SECTION`` allow-list); ``ONE``
-    yields one row per article. Columns are the section fields in their
-    resolved order; values are baked from ``value_map`` (already-resolved
-    scalars — §5.3). The ``MODEL_CONTAINER`` and field-less sections are
-    skipped (nothing to project).
+    The record axis is selected by ROLE first, then cardinality — mirroring
+    ``matrix._resolve_instance_id``:
+
+      * ``MODEL_SECTION`` fans out one row per model instance
+        (``article.model_instances``) regardless of its own cardinality.
+        Production model sections are ``cardinality='one'``; the N-model
+        fan-out is always sourced from ``model_instances``, never from
+        ``section_instances`` (spec §5.2).
+      * non-model ``cardinality==MANY`` fans out one row per instance
+        (``section_instances``).
+      * non-model ``cardinality==ONE`` yields one row per article.
+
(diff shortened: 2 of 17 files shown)


raphaelfh:
## Claude review
**Verdict:** LGTM with nits
**Scope:** Post-merge adversarial fix — silent data loss in MODEL_SECTION tidy-table export (role-first fan-out), crash in `usePageHandle` outside a `ViewerProvider`, and two confused-deputy BOLAs in run-resolution and template-loading, each with a targeted regression test.

### Findings

- **minor** `frontend/pdf-viewer/hooks/usePageHandle.ts:10` — `NO_PROVIDER_STORE` is a module-level Zustand store object; the comment calls it "never mutated" but nothing mechanically enforces that. It is unexported and only subscribed read-only via `useStore(..., s => s.document)`, so the practical risk is negligible. A one-liner `// invariant: state is never written; document stays null` where the store is used (instead of in the declaration comment) makes it easier to audit at call sites.

- **minor** `backend/app/services/extraction_run_read_service.py:534` — `project_id` is now a required keyword argument on `resolve_form_runs`. The diff updates the one visible call site (`articles.py`). Since this is a public module-level function it could have other callers (e.g., direct service-layer tests) that would fail with `TypeError` at runtime rather than at import time. A quick `grep -r "resolve_form_runs" backend/` before merge confirms nothing was missed. CI on a clean schema is the gate, and the PR discloses the local block, so not blocking — just worth the 10-second check.

### Incident-class sweep

| Class | Status |
|---|---|
| BOLA | **fixed** — `resolve_form_runs` and `_load_active_template_version` both now scoped by `project_id`; positive-control tests confirm legitimate same-project access still resolves |
| TOCTOU | ok — no run-state transitions touched |
| Error swallowing | ok — `NotFoundError` raised cleanly on cross-project template load; no bare except / swallowed catch |
| ApiResponse envelope | ok — endpoint returns `ApiResponse.success(refs, ...)` unchanged |
| TanStack cache | n/a — no query-key or invalidation changes |
| Schema drift | n/a — no SQLAlchemy models or migrations |
| RLS | n/a — no policy changes |

---

PR #382, opened 2026-06-23
feat(reader): pro markdown rendering, markdown-first citation locate, anchored suggestion popover

raphaelfh:
## What & why

The document reader rendered LlamaParse markdown as **raw text** — clinical tables showed as `| … |` pipes, flow charts as bare `graph TD …`, and `<page_number>` tags verbatim. The AI **Suggestion details** view was a full-screen modal that covered the very document it asked you to inspect, and the PDF-bbox citation locator was unreliable. (See the before-screenshots in the originating issue.)

## Changes (frontend-only — artifacts cleaned at render time, no backend/migration)

**1. Professional markdown rendering** — new `frontend/pdf-viewer/markdown/`:
- `react-markdown` + `remark-gfm` (tables, lists, strikethrough, autolinks)
- `remark-math` + `rehype-katex` (formulas)
- lazy, fail-safe **Mermaid** for flow diagrams (code-split out of the main bundle; falls back to source on parse failure)
- `<page_number>` artifacts stripped; raw HTML inert (XSS-safe by default); tuned Tailwind Typography for the dark minimalist theme

**2. Markdown-first citation locate** — a `readerLocate` request on the shared viewer store (`locateInReader`) switches to reader mode and the Reader scrolls to + flashes the block whose text matches the evidence quote (markdown-syntax tolerant). Replaces the fragile bbox jump. The document panel auto-opens if collapsed.

**3. Detail modal → anchored, non-modal `Popover`** — the document stays visible/interactive while locating; **Locate in document** closes the popover and flashes the cited passage. Also fixes a hard-coded Portuguese `"Justificativa"` string (now an English copy key).

## Verification
- `npm run test:run` — **836 passed**; `tsc` clean; `npm run build` clean; `eslint` clean.
- Visual loop at desktop (1280) **and** narrow (390): tables/math/mermaid render professionally; locate flash + anchored popover match the minimalist target.
- `/simplify` applied (shared `subscribeReaderLocate` helper, `blocksRef` stable subscription, precomputed match normalization).

## Follow-up (out of scope, flagged)
This makes the old PDF-bbox citation stack (`useCitationHighlight`, `citationsService`, `CitationOverlay`, `/articles/{id}/citations`) and the threaded `articleId` popover prop unreachable from the UI. A deletion sweep is tracked separately rather than widening this PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #383, opened 2026-06-23
Promote dev → main: pro markdown reader + markdown-first citation locate + anchored popover (#382)

raphaelfh:
Production promotion of #382.

**Frontend-only** (no backend/migration). The document reader now renders LlamaParse markdown professionally (GFM tables, KaTeX math, lazy/fail-safe Mermaid flow diagrams; `<page_number>` artifacts stripped; XSS-safe). Citation locate is markdown-first (scroll + flash the cited block). The AI suggestion detail view is an anchored, non-modal popover that no longer covers the document.

Verified on dev: 836 vitest + tsc + build + eslint + Architectural Fitness all green; visual loop (desktop + 390px) confirmed.


raphaelfh:
## Claude review
**Verdict:** LGTM with nits
**Scope:** Frontend-only: replaces the flat per-block `<p>` dump with a full `react-markdown` rendering pipeline (GFM tables, KaTeX math, lazy Mermaid diagrams, `@tailwindcss/typography` prose styling); rewires citation locate to a nonce-based viewer store subscription (drops the `useArticleCitations` fetch path); converts the AI suggestion detail view from a full-screen Dialog to an anchored non-modal Popover.

### Findings

- **nit** `frontend/pdf-viewer/markdown/Mermaid.tsx:53–58` — Mermaid parse/render errors are silently discarded. The fallback to a code-block is correct, but a `console.warn(err)` in the catch would make malformed LlamaParse diagrams debuggable without adding observability overhead. Not a correctness issue.

- **nit** `frontend/pdf-viewer/markdown/Mermaid.tsx:13–30` (module-level `initialized` singleton) — Mermaid's dark/light theme is captured at the time of the first diagram render. A user who opens a document before switching themes, then switches, will see mismatched diagram colours until page reload. Known Mermaid limitation; worth a short `// Note: re-init requires page reload` comment so the next reader isn't surprised.

- **nit** `docs/adr/0013-dual-tier-markdown-representation.md` — ADR-0013 prescribes `rehype-sanitize` (strict allowlist) as defense in depth. The PR instead relies on the absence of `rehype-raw`, which is stricter (HTML never renders at all) and is well-tested (`MarkdownContent.test.tsx:33–38`). The ADR also mentions server-side `nh3`/`bleach` sanitization at write time; that's deferred as this is frontend-only. No security gap, but updating ADR-0013's status from `proposed` to `accepted` (or adding an implementation note noting the `no-rehype-raw` approach) would keep the ADR accurate.

### Incident-class sweep

| Class | Status |
|---|---|
| BOLA | n/a — frontend-only, no new endpoints |
| Run-state TOCTOU | n/a — no backend state mutations |
| Swallowed errors | ok — Mermaid fail-safe is intentional; no empty `.catch()` on important paths |
| ApiResponse envelope | ok — citation fetch removed cleanly; no new API calls added |
| TanStack cache | ok — no new mutations; `useArticleCitations` removed from Popover correctly |
| Schema drift | ok — API Contract CI green; no hand-mirrored types added |
| RLS | n/a — no migrations |
| XSS | ok — no `rehype-raw`; `dangerouslySetInnerHTML` only via `mermaid.render()` with `securityLevel:'strict'`; XSS test in place |

---

PR #387, opened 2026-06-24
chore(extraction): remove orphaned PDF-bbox citation stack

raphaelfh:
## Why

[#382](https://github.com/raphaelfh/prumo/pull/382) moved AI-evidence citation locating to a markdown-first reader (`useReaderLocate` + the viewer-store `readerLocate` slice). That made the old **PDF-bbox citation system unreachable from the UI**. This is a focused, behavior-preserving removal of that dead stack.

## What changed

**Frontend — delete (each verified orphaned by grep):**
- `useCitationHighlight.ts`, `services/citationsService.ts`, `hooks/articles/useArticleCitations.ts`
- `pdf-viewer/primitives/CitationOverlay.tsx`, `CitationLiveRegion.tsx`, `pdf-viewer/core/citation.ts` (+ all their tests)

**Frontend — edit:**
- viewer store: removed the `citations`/`activeCitationId` slice + `addCitation`/`removeCitation`/`clearCitations`/`setActiveCitation` (`core/{state,store,types}.ts`); kept the live `readerLocate`/`locateInReader` slice
- `Viewer.tsx` / `PrumoPdfViewer.tsx`: removed the now write-orphaned `<CitationOverlay>` / `<CitationLiveRegion>` renders
- `ExtractionPDFPanel.tsx`: dropped the now-inert `clearCitations()` on document switch
- removed orphaned copy keys (`citationHighlightLabel`, `citationJumpAnnouncement`), the unused `articleKeys.citations` query-key, and stale doc/comment refs (`pdf-viewer/README.md`, `DocumentSwitcher.tsx`, `usePageHandle.ts`)

**Frontend — `articleId` prop leg (narrowly scoped):** stripped only the leg that fed the dead popover — `AISuggestionDetailsPopover` → `AISuggestionDisplay` → `FieldInput` → `InstanceCard`, plus `QASectionAccordion`. **Kept** `articleId` everywhere it has a real use: `ExtractionFormView`/`ModelSection`/`SectionAccordion` (per-section/model AI extraction), `ExtractionPDFPanel`, `ParseStatusControl`, `QualityAssessmentFullScreen`. Verified both directions with `tsc`.

**Backend (scope narrowed — see Risk):**
- delete the now frontend-unreachable `GET /api/v1/articles/{id}/citations` **endpoint** + its router registration + `test_citations_endpoint.py`
[...]

PR #386, opened 2026-06-23
feat(extraction): A1 — block-markdown LLM input + windowing

raphaelfh:
## A1 — LLM consumes a block-derived markdown projection, windowed

Finishes ADR-0011's grounded-extraction half: the LLM is now fed a **budgeted markdown projection of `article_text_blocks`** instead of `pypdf` text hard-truncated at 15,000 chars. Implements §A1 of the [stabilization spec](docs/superpowers/specs/2026-06-23-extraction-pipeline-stabilization-design.md). Plan: [2026-06-23-extraction-a1-block-input.md](docs/superpowers/plans/2026-06-23-extraction-a1-block-input.md).

### What changed
- **`render_blocks_to_markdown(blocks)`** (pure, in `infrastructure/parsing/base.py`) — the single canonical block→GFM codepath (ADR-0013 free tier): `##` headings, `-` lists, contiguous same-page `table_cell` runs → GFM tables, chrome suppressed, reading order.
- **Reused #325's tested assembler** rather than building anew: relocated `extraction_block_assembler.py` → **`app/llm/assembler.py`** and made `_serialize_section` delegate to the renderer (one table codepath). Its ~24 behavior tests carry over unchanged.
- **`assemble_for_model(blocks, *, model_name, budget_tokens)`** + internal **`AssemblyInfo`** (`total_blocks`/`included_blocks`/`truncated`/`est_tokens`; tiktoken for OpenAI, char/4 heuristic + documented Anthropic skew). `LLM_ASSEMBLY_BUDGET_TOKENS=96_000`.
- **Deleted `MAX_PDF_CHARS`** truncation at all three render sites (`section_extraction`, `model_identification`, `quality_assessment`).
- **`build_prompt_input`** (`services/extraction_prompt_input.py`) fetches blocks **once per run**, assembles **once**, threads the markdown through the entity-type loop, reuses the blocks for evidence anchoring, and routes the **no-blocks pypdf fallback through the same budgeted assembler** (no path sends unbounded text). Wired into both `section_extraction_service` and `model_extraction_service`.

### Tests / gates
- Full-chain integration test (#10): post-15k block → `assemble` (no truncation) → `build_anchor` maps back to the correct block → `citation_read_service` round-trip.
- Direct-coroutine unit tests for both service paths (blocks-assemble vs pypdf-fallback) — the diff-cover ASGI blind spot.
[...]

PR #385, opened 2026-06-23
feat(extraction): centralize LLM model/provider + extractor reliability (Phase 1)

raphaelfh:
## Extraction pipeline stabilization — Phase 1 (C1 + extractor reliability)

First of three incremental PRs from the extraction-stabilization effort (spec + plan + adversarial review included on this branch). **Migration-free.** The live default is unchanged (`LLM_PROVIDER=openai`, `gpt-4o-mini`) — no behaviour or cost change until config flips.

### Decisions (recorded in the spec on this branch)
- **A1** — LLM consumes block-derived markdown + windowing (Phase 2).
- **B1** — text-first citations; delete the orphaned bbox stack (Phase 3).
- **C1** — keep `pydantic-ai`; one configurable model/provider; Claude selectable (BYOK). **This PR delivers C1 + the reliability bugs.**

### What changed
- **Model/provider config centralized**: one `LLM_PROVIDER` / `LLM_DEFAULT_MODEL` / `LLM_TIMEOUT_SECONDS`; the **11 hardcoded `gpt-4o-mini` literals** across the extraction pipeline are gone; **dead `OPENAI_DEFAULT_MODEL` removed**.
- **`build_model(provider, model_name, *, api_key)`** — OpenAI branch unchanged + lazy **Anthropic** branch (BYOK-only, no global key); `pydantic-ai-slim[anthropic]` extra. Endpoints + Celery tasks resolve the key for `settings.LLM_PROVIDER`.
- **Provider-aware structured output**: `NativeOutput` (OpenAI) vs `ToolOutput` (Anthropic, detected via `model.system`) — Claude has no `response_format`. Plus a **client-level timeout** that actually aborts the request.
- **Celery retry**: typed transient/permanent taxonomy → exponential backoff + jitter on transient, **fail-fast** on permanent (missing key/PDF/template, `UsageLimitExceeded`).
- **Batch fails closed**: an all-sections-failed batch raises `BatchAllSectionsFailed` (→ run **FAILED**, persisted) instead of reporting a false success; partial failures still complete and surface the failed list.
- **Schema fails closed** on duplicate template field names (`SchemaBuildError`) instead of silently dropping data.
- **Disproven premises guarded, not patched**: `.format()` injection (proven safe), late key validation (already early), the `#bug` marker (design-intent) — locked with design-stability assertions.

### Review
[...]

PR #389, opened 2026-06-24
promote: dev → main (consensus Phase A + citation cleanup + A1 + extractor reliability)

raphaelfh:
## Promote `dev` → `main` (production release)

Brings `main` up to `dev` (4 PRs, all CI-green on `dev`):

- **#388** feat(consensus): reconciliation-by-state view + multi-reviewer adequacy (Phase A) — frontend-only, adversarially reviewed, design-review passed.
- **#387** chore(extraction): remove orphaned PDF-bbox citation stack — FE+BE cleanup.
- **#386** feat(extraction): A1 — block-markdown LLM input + windowing.
- **#385** feat(extraction): centralize LLM model/provider + extractor reliability (Phase 1).

Merge-commit promotion (not squash/ff) per the dev→main convention. Railway (web+worker) deploys from `main`; Vercel redeploys the frontend.


raphaelfh:
## Claude review
**Verdict:** LGTM with nits
**Scope:** dev→main promotion of 4 PRs — provider-agnostic LLM layer (Anthropic/Claude + OpenAI), budgeted block-markdown assembler replacing ad-hoc truncation, Celery retry classification + `BatchAllSectionsFailed` error handling, consensus reconciliation panel refactor to structured buckets. No migrations, no new tables, no auth-surface changes.

> **Diff exceeded 4,000 lines (11,393 lines, 106 files).** Review scoped to `backend/app/`, `frontend/{services,hooks,components}/`, `supabase/migrations/`, and `backend/alembic/` per the automated review protocol.

---

### Findings

- **Important** `backend/app/services/section_extraction_service.py` ~line 1339 — When `pdf_text` is pre-supplied by the caller and `self._run_anchor_blocks` is empty, `_assemble_prompt_text` is called a second time but its return value is discarded. The intent is to populate `self._run_anchor_blocks` for evidence anchoring, but the anchor blocks are now built from a second assembly run while the prompt was generated from the caller-supplied text. These two inputs can diverge (different block selection, different windowing), so evidence anchors may not correspond to what the LLM actually saw. Fix: either always prefer the assembled text over the caller-supplied `pdf_text`, or extract block-fetching into a separate method that doesn't re-run the full assembler.

- **Important** `backend/app/worker/tasks/extraction_tasks.py` ~line 115 — `BatchAllSectionsFailed` is classified as permanent (non-transient) by `is_transient_llm_error`, so Celery propagates it without retry. The key question is whether `rollback_and_fail` in `RunLifecycleService` commits the `FAILED` state to the DB before re-raising. If it only rolls back the transaction and then raises, the run stays in its pre-failure state and callers see a zombie run. `test_all_sections_failed_commits_failed_status` covers the HTTP path; confirm the same commit path is exercised via the Celery task. Fix: add a Celery-path integration test or grep for `commit()` in `rollback_and_fail` to confirm.

- **Minor** `frontend/hooks/extraction/useConsensusReconciliation.ts` ~line 484 — `useProjectMembers(projectId ?? '')` is called unconditionally. When `projectId` is `undefined`, an empty-string project ID is passed to the hook. If `useProjectMembers` does not gate on `enabled: !!projectId`, this fires a real network call with an empty path. Fix: `useProjectMembers(projectId ?? '', { enabled: !!projectId })`.

- **Minor** `backend/app/llm/extractor.py` ~line 821 — Provider detection uses `getattr(model, "system", "") == "anthropic"`, which relies on an undocumented pydantic-ai internal attribute. A pydantic-ai minor release could rename it and silently fall back to `NativeOutput` for Anthropic models. Not a current bug, but worth an `isinstance` check or a version-pinned assertion in CI.

- **Minor** `backend/app/llm/assembler.py` ~line 686 — The char-budget heuristic (`budget_tokens * 4`) assumes ~4 chars/token. For non-ASCII scientific papers (Greek symbols, CJK references, math notation), the actual ratio is lower, meaning the assembled document may exceed the token budget silently. Not a hard failure, but monitoring `est_tokens` in Logfire against the 96 k ceiling is advised before Anthropic support goes to production.

---

### Incident-class sweep

| Class | Status |
|---|---|
| BOLA | ok — deleted citations endpoint had `ensure_project_member` wired; no new endpoints without auth |
| TOCTOU | finding: see Finding 2 above — verify `rollback_and_fail` commits before raising in Celery path |
| Swallowed errors | ok — old bare `except Exception: _anchor_blocks = []` was removed in this PR; no new swallowing introduced |
| ApiResponse envelope | ok — no new bare payload returns; deleted endpoint used envelope correctly |
| TanStack cache | ok — `useArticleCitations` key deleted with its endpoint; no orphaned invalidations; `useConsensusReconciliation` is pure computation |
| Schema drift | n/a — no Pydantic/SQLAlchemy column changes; `AssemblyInfo` explicitly excluded from response models |
| RLS | n/a — no new tables or migrations |

---

PR #391, opened 2026-06-24
fix(parsing): disable OCR in Docling parser (RapidOCR PP-OCRv6 crash)

raphaelfh:
## Why

The self-hosted **Docling** fallback parser crashes on **every** PDF that reaches it with `ValueError: Unsupported configuration: torch.PP-OCRv6.det.small`. Three prod `article_files` are stuck at `parse_failed` (most recent today 12:15 UTC), retrying every minute.

Root cause: docling's default `do_ocr=True` loads RapidOCR. **RapidOCR 3.9** (released ~2026-06-23 — matches the first failure at 06-23 19:26) promoted a **PP-OCRv6** detection model whose **torch** backend has no registered model URL. The slim worker ships **no `onnxruntime`**, so RapidOCR falls back to its torch engine and raises at pipeline init, before any page is read. `pyproject` pins `docling>=2.0.0` and the Dockerfile installs via `uv pip install -e .` (ignores `uv.lock`), so the transitive rapidocr drifted to 3.9 on a rebuild — which is why prod broke while local (rapidocr 3.8.4 → PP-OCRv4) still works.

## What changed

- `backend/app/infrastructure/parsing/docling_parser.py`: new `_pdf_pipeline_options()` builds the converter with **`do_ocr=False`**, `do_table_structure=True` (`TableFormerMode.FAST` + `do_cell_matching=True`), and the accelerator pinned to CPU. `parse()` now builds the `DocumentConverter` from these options instead of a bare `DocumentConverter()`.
- `backend/tests/unit/test_docling_parser_options.py`: regression test asserting OCR is off, tables stay on (FAST + cell matching), CPU accelerator.

Why `do_ocr=False` is correct and sufficient: born-digital scientific PDFs carry an embedded text layer, so OCR adds nothing. Confirmed from docling source that `from rapidocr import RapidOCR` lives inside `if self.enabled:` — with OCR off, **RapidOCR is never imported**, so the rapidocr version drift can't crash us regardless of which version a rebuild resolves. bbox + table structure are preserved (layout + TableFormer stay on), so evidence-anchoring/hybrid citations and table extraction are unaffected.

## Risk

Low, isolated to the Docling adapter (no endpoint / DB / migration / API-envelope surface). Behavioural change: a genuinely **scanned** (image-only) PDF hitting the self-hosted path now yields no text → `parse_failed` (the *same* terminal state as today, since it currently crashes for everyone). Default path is LlamaParse cloud; Docling is the fallback.

## Test plan

- [x] New regression test red→green (`tests/unit/test_docling_parser_options.py`)
- [x] `tests/integration/test_docling_parser.py` passes (real parse, OCR off)
- [x] Confirmed `rapidocr` not in `sys.modules` after a `DoclingParser().parse(...)`
- [x] Full unit suite: 1580 passed
- [x] `ruff check app tests` clean; `ruff format --check` clean
- [ ] Full `make test-backend` integration suite — deferred to CI (needs local Supabase Docker; change has no DB surface)

## Out of scope (follow-up: "harden next")

Build reproducibility + cost: switch Dockerfile to `uv sync --frozen` (honor the lock), pin torch/torchvision to the PyTorch **CPU** index (drops ~1GB+ of CUDA wheels — 74 nvidia refs in the lock), and prefetch only layout+TableFormer models at build time with `DOCLING_ARTIFACTS_PATH` + `HF_HUB_OFFLINE` (no runtime downloads). Tracked for PR 2.

## Post-deploy

After dev→main promotion + Railway redeploy, re-parse the 3 `parse_failed` files (and the pending backlog) via the existing re-parse affordance.

### DIFF
--- a/backend/app/infrastructure/parsing/docling_parser.py
+++ b/backend/app/infrastructure/parsing/docling_parser.py
@@ -5,6 +5,14 @@ lazy-imported inside parse() so app boot and non-parsing tests stay light.
 Maps docling DocItem labels onto the closed block_type set, reads bbox from
 each item's prov, and emits ParsedBlock with char offsets as 0 placeholders
 (DocumentParsingService assigns real offsets).
+
+OCR is disabled: our inputs are born-digital scientific PDFs with an embedded
+text layer, so OCR adds nothing. docling's default ``do_ocr=True`` loads
+RapidOCR, whose torch backend crashes on the PP-OCRv6 default it adopted in
+3.9 ("Unsupported configuration: torch.PP-OCRv6.det.small") when onnxruntime
+is absent. Disabling OCR skips the RapidOCR import entirely. Table structure
+stays on (FAST + cell matching: cell text comes verbatim from the PDF text
+layer) since scientific tables are load-bearing for extraction.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


raphaelfh:
## Claude review
**Verdict:** LGTM
**Scope:** Disables OCR in the self-hosted Docling fallback parser by extracting `_pdf_pipeline_options()` that sets `do_ocr=False`, `do_table_structure=True` (FAST + cell matching), and CPU accelerator — fixing a prod crash caused by RapidOCR 3.9's PP-OCRv6 torch backend being unresolvable on the slim worker. Regression test added.

### Findings

### Incident-class sweep

| Class | Status |
|---|---|
| Swallowed errors | ok — exceptions from `converter.convert()` propagate naturally; no bare excepts added |
| ApiResponse envelope | n/a — no API surface |
| TanStack cache | n/a — backend-only change |
| Schema drift | n/a — no models/migrations |
| RLS | n/a — no tables or migrations |

**Notes:**
- `converter` is constructed per `parse()` call, same as the old `DocumentConverter().convert(...)` pattern — no regression, no change in per-call overhead.
- Lazy imports inside `_pdf_pipeline_options()` keep app boot light — correct pattern.
- Test correctly skips when docling is absent and guards the exact regression case (`opts.do_ocr is False`). Missing `assert "rapidocr" not in sys.modules` (mentioned in the test plan as manually verified), but the mechanism is well-documented and the configuration assertion is sufficient.
- The `uv sync --frozen` + CUDA wheel + model prefetch hardening tracked for follow-up PR is the right call — this fix is scoped correctly.

---

PR #394, opened 2026-06-24
chore(parsing): CPU-only torch + prefetched Docling models + lean image build

raphaelfh:
## Why

Follow-up hardening to [#391](https://github.com/raphaelfh/prumo/pull/391) (which disabled OCR to fix the RapidOCR PP-OCRv6 crash). Two cost/reliability problems remained in the worker image:

1. The CPU-only worker pulled the **full CUDA torch stack** (~1GB of `nvidia-*`/`triton`) as dead weight — 74 nvidia refs in the resolution.
2. Docling downloaded its models from HF Hub at **first parse** (slow, network-fragile on the worker).

## What changed (`backend/Dockerfile`, `pyproject.toml`, `uv.lock`, `.dockerignore`)

- **CPU-only torch**: `uv pip install --torch-backend=cpu` resolves torch/torchvision from the PyTorch CPU index → `torch 2.12.1+cpu`, dropping the entire CUDA stack.
- **Build-time model prefetch**: download *only* layout + TableFormer into `/opt/docling-models` (skip OCR + the ~500MB code-formula model), then `DOCLING_ARTIFACTS_PATH` + `HF_HUB_OFFLINE=1` + `TRANSFORMERS_OFFLINE=1` so the worker **never** downloads at runtime.
- **Pin `docling==2.104.0`**: the parser stack is drift-sensitive (an unpinned rapidocr 3.9/PP-OCRv6 bump broke prod). OCR is disabled in #391 so rapidocr drift is moot, but pinning docling stops its own API/model drift.
- **`backend/.dockerignore`**: smaller, secret-free build context.

## Risk

Build-system only; no app code changes. Image build is the risk surface — validated below. Must deploy **with or after #391** (this image prefetches no OCR models and forbids runtime downloads; OCR-on code would fail offline). Both go to `main` in the same promotion.

## Test plan / evidence (validated locally on `linux/amd64`, Railway's arch)

- [x] `docker build --platform linux/amd64` succeeds (image 1.62GB)
- [x] `torch 2.12.1+cpu`, `torchvision 0.27.1+cpu`, **zero** nvidia/triton packages in the image
- [x] **Offline parse** (`docker run --network none`) succeeds from prefetched models with `do_ocr=False` — 4 blocks, text present
- [x] `uv lock --check` consistent; `ruff check` clean
- [ ] CI **Backend Docker Build** rebuilds the committed files (authoritative gate)

## Out of scope / notes

- **Full `uv sync --frozen` deferred**: uv 0.10.9 cannot combine a frozen lock with CPU-torch selection (`--torch-backend` is `uv pip install`-only; `uv lock`/`uv sync` ignore it and mis-resolve torchvision to a wheel-less local build). docling is pinned as the targeted drift guard; #391 already fixes the rapidocr drift that caused the outage.
- The `apt-get install libgl1 libxcb1 …` block is kept conservatively; with OCR off it is likely removable, but only as a separate tested cleanup (confirm no dep imports `cv2`).

### DIFF
--- a/backend/.dockerignore
+++ b/backend/.dockerignore
@@ -1,0 +1,17 @@
+# Keep the build context small and secret-free. The Dockerfile only needs
+# pyproject.toml, uv.lock, README.md, app/, alembic/, alembic.ini.
+.venv
+venv
+.env
+.env.*
+__pycache__/
+*.pyc
+*.pyo
+.pytest_cache/
+.ruff_cache/
+.mypy_cache/
+.coverage
+htmlcov/
+tests/
+.git/
+*.egg-info/
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -12,20 +12,27 @@ RUN pip install --no-cache-dir uv
 COPY pyproject.toml README.md ./
 COPY app ./app
 
-# Install dependencies (editable so the runtime stage can reuse site-packages)
-RUN uv pip install --system --no-cache -e .
+# Install dependencies into the system prefix (the runtime stage copies it).
+# --torch-backend=cpu makes uv resolve torch/torchvision from the PyTorch CPU
+# wheel index, dropping the ~1GB CUDA stack (nvidia-*/triton) that is dead
+# weight on the CPU-only worker.
+RUN uv pip install --system --no-cache --torch-backend=cpu -e .
+
+# Prefetch ONLY the models the OCR-disabled standard pipeline needs (layout +
+# TableFormer) into the image, so the worker NEVER downloads at runtime. OCR
+# (RapidOCR) is disabled in the parser, and the code-formula / picture models
+# are unused — skip them to keep the image lean.
+RUN python -c "from pathlib import Path; from docling.utils.model_downloader import download_models; download_models(output_dir=Path('/opt/docling-models'), with_layout=True, with_tableformer=True, with_code_formula=False, with_picture_classifier=False, with_rapidocr=False, with_easyocr=False)"
 
 # =================== RUNTIME STAGE ===================
 FROM python:3.12-slim as runtime
 
 WORKDIR /app
 
-# System libraries for the self-hosted Docling fallback parser. Docling pulls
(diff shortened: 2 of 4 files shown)


raphaelfh:
## Claude review
**Verdict:** LGTM with nits
**Scope:** Builds a CPU-only worker image — swaps CUDA torch for the PyTorch CPU wheel index, prefetches Docling layout + TableFormer models at build time, sets `HF_HUB_OFFLINE=1` / `TRANSFORMERS_OFFLINE=1` so the worker never touches the network at runtime, pins `docling==2.104.0` to guard against API/model drift. Build-system only; no app logic changed.

### Findings

- **nit** `backend/Dockerfile:21` — The `download_models` kwargs (e.g. `with_picture_classifier`, `with_easyocr`) are embedded directly in the `RUN python -c "…"` one-liner. While the `docling==2.104.0` pin freezes the API for now, a future version bump may rename a kwarg and the build fails with an opaque Python error rather than a clear diff. A small `scripts/prefetch_docling_models.py` would make the download step greppable and independently testable.

- **nit** `backend/Dockerfile:14` — The comment above the `uv pip install --torch-backend=cpu` line explains the CPU-wheel selection, but not that it also bypasses `uv.lock` (already noted in the PR body). A one-line addition — e.g. `# NB: uv pip install does not read uv.lock; see PR #394 for why uv sync --frozen is deferred` — prevents a future maintainer from "fixing" it by switching to `uv sync --frozen` without understanding the torch-backend constraint.

- **consideration** `HF_HUB_OFFLINE=1` is set image-wide with no docs trail. Any future HF-backed model added to the app will fail with a cryptic offline error at runtime rather than a build-time signal. The PR body explains this well; a short `docs/adr/` entry would make the constraint discoverable for future devs who don't read PR history.

### Incident-class sweep

| Class | Status |
|---|---|
| BOLA | n/a — no endpoint changes |
| Run-state TOCTOU | n/a — no state machine code |
| Error swallowing | n/a — no app Python changed |
| ApiResponse envelope | n/a — no API changes |
| Schema drift | n/a — no model/migration/type changes |
| RLS | n/a — no Supabase migrations |

---

PR #425, opened 2026-06-28
fix(extraction): format %<->fraction candidates with .12g to avoid float-repr noise

raphaelfh:
## Why

The entailment gate's deterministic numeric check (`backend/app/llm/value_support.py`) marks legitimate numeric citations as `unsupported` (yellow) whenever the extracted value and the source text use opposite percentage/fraction forms — e.g. value `11.8%` vs source `0.118`. Root cause: the `%<->fraction` equivalence candidates are built with `str(f / 100)` / `str(f * 100)`, and Python's default float repr leaks binary noise:

- `str(11.8 / 100)` -> `"0.11800000000000001"` (not `"0.118"`)
- `str(0.118 * 100)` -> `"11.799999999999999"` (not `"11.8"`)

The noisy candidate never matches the clean decimal in the text, so the gate returns `False`. Conservative failure mode — only **false negatives** (a valid citation suppressed), never a false verification.

## What changed

- [value_support.py:85](backend/app/llm/value_support.py#L85): build the `%<->fraction` candidates with `format(f, ".12g")` instead of `str(f)`. 12 significant digits drops the binary tail (`0.118`, `0.007`, `11.8`) while `_canon` stays the *conditional* strip (`if "." in n`) so an integer candidate like `"500"` is not mangled to `"5"`.
- [test_value_support.py](backend/tests/unit/test_value_support.py): regression test for both directions of the float-noise bug, plus a guard test pinning the conditional-strip behavior.

## Risk

Low. Pure function, no IO, no schema/migration. The change only *adds* matches that should always have matched (cleaner candidate strings); it cannot create a false positive because the deterministic check is one half of the entailment gate (the LLM judge still runs). Pre-existing bug since #416 — **not** a regression of the thousands-separator work, whose tests stay green.

## Test plan

- [x] `make lint-backend` clean (ruff check + format, all 497 backend files)
- [x] Backend unit suite: `uv run pytest tests/unit/` — 1657 passed
- [x] New regression test fails on `dev`, passes after the fix (verified red->green)
- [ ] CI runs the full integration suite

## Out of scope

None.

### DIFF
--- a/backend/app/llm/value_support.py
+++ b/backend/app/llm/value_support.py
@@ -82,8 +82,11 @@ def _candidates(value: str) -> set[str]:
                 f = float(n)
             except ValueError:
                 continue
-            out.add(_canon(str(f / 100)))  # 12.5 -> 0.125
-            out.add(_canon(str(f * 100)))  # 0.125 -> 12.5
+            # `.12g` keeps 12 significant digits without leaking binary float
+            # noise (`str(11.8 / 100)` -> "0.11800000000000001"), so the clean
+            # decimal in the source text still matches.
+            out.add(_canon(format(f / 100, ".12g")))  # 12.5 -> 0.125
+            out.add(_canon(format(f * 100, ".12g")))  # 0.125 -> 12.5
     return {c for c in out if c}
 
 
--- a/backend/tests/unit/test_value_support.py
+++ b/backend/tests/unit/test_value_support.py
@@ -43,3 +43,30 @@ def test_comma_decimal_not_misread_as_grouping():
     # Locale comma-decimal ("11,8" == 11.8) must read as a decimal, not 118.
     assert numeric_value_supported("11,8", "increased by 11.8 points")
     assert not numeric_value_supported("11,8", "a total of 118 events")
+
+
+def test_percent_fraction_forms_match_despite_float_noise():
+    """Regression: the %<->fraction candidates must not leak binary float noise.
+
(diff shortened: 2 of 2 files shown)


raphaelfh:
## Claude review
**Verdict:** LGTM
**Scope:** Fixes a false-negative in the deterministic numeric entailment gate by replacing `str(f / 100)` / `str(f * 100)` with `format(f, ".12g")` so binary float noise (`0.11800000000000001`) no longer prevents a clean decimal in the source text from matching.

### Findings

### Incident-class sweep

| Class | Status |
|---|---|
| BOLA | n/a — pure function, no endpoints |
| Swallowed errors | n/a — no exception handling in diff |
| ApiResponse envelope | n/a — no API layer touched |
| TanStack cache | n/a — backend-only change |
| Schema drift | n/a — no models or migrations |
| RLS | n/a — no DB access |

**Fix correctness:** `.12g` gives 12 significant digits and `g`-semantics trailing-zero suppression. `format(11.8/100, ".12g")` → `"0.118"`, `format(500.0, ".12g")` → `"500"` (no decimal point, so `_canon`'s conditional strip leaves integers intact — the guard test pins this). Scientific notation behaviour for extreme values is unchanged from the old `str()` path (both use the same < −4 exponent threshold).

**Tests:** Two well-named regression tests — one for the reported bug (both directions of the %↔fraction conversion), one guarding the tempting-wrong-fix of unconditional trailing-zero stripping. No flaky patterns.

---

PR #443, opened 2026-06-29
feat(extraction): unified AI-suggestion review + run provenance + no-information proposals

raphaelfh:
## What

Replaces the broken split AI-suggestion popovers with one **selection-based review surface**, surfaces **run provenance** (how each suggestion was generated), and fixes the silently-dropped **"no information"** extraction outcome. Establishes *transparency & traceability of AI-assisted decisions* as a constitution pillar (§IX).

Implements [`docs/superpowers/plans/2026-06-29-suggestion-review-and-provenance.md`](docs/superpowers/plans/2026-06-29-suggestion-review-and-provenance.md) (spec + adversarially-reconciled plan in-tree).

## Backend
- **No-information proposals** (`_create_suggestions`): abstention / `not_found` now records a first-class proposal `{value: null}` (rationale kept, no misleading 0% confidence) instead of a silent `continue`. `record_proposal` stays run-scoped idempotent (no dup on re-run).
- **Run provenance snapshot** into `extraction_runs.results['provenance']` (no migration): provider, model, strategy, prompt version + the prompt actually sent, params (single-source `LLM_TEMPERATURE`/`OUTPUT_RETRIES_DEFAULT` constants), tokens, ran-by user.
- **API exposure**: `provenance` on the suggestions + history items, attached via one batched run query (no N+1). Generated API types regenerated.

## Frontend
- **`AISuggestionReviewPopover`** — one surface (replaces history + details popovers): versions grouped by run, pure-selection model (one *Selected*, others *Use this version*), selected version expands to `RunProvenanceDisclosure` + cited evidence with reader-locate; null → quiet **No information found** card; pinned footer keeps Clear reachable.
- **`RunProvenanceDisclosure`** — extensible, data-driven (known-field registry + generic fallback for unknown keys → new backend fields render with zero FE change).
- **`selectSuggestion`** accept-by-proposal-id (carries the chosen version's value + confidence); `AIPopoverShell` gains a pinned footer slot.
- Inline strip shows a **quiet** "No information found" (not a loud `(empty) · 0%` accept/reject); `aiPendingCount` excludes no-info.
- Removed `AISuggestionHistoryPopover`, `AISuggestionDetailsPopover`, and the now-dead `AISuggestionInline`.

## Verification
- Frontend: `lint`, `typecheck`, **924 vitest** (incl. new mapping/disclosure/hook/popover/Display/autosave-R7 tests), React-Compiler `build`, data-path + react-query-key fitness — all green.
- Backend: new unit tests for the provenance **write-path** (`_provenance_with_tokens` token-merge + None branch) + no-info proposal emission; integration (`test_suggestion_read`) via CI Backend Tests.
- **Adversarial 5-lens review** (layering / security / correctness / no-info+QA-prefill / coverage) with per-finding verification: security clean (controls unchanged, provenance carries no secrets/PHI); 3 confirmed findings (quiet no-info strip, reopen-highlight, provenance write-path coverage) fixed; `/simplify` pass applied.


raphaelfh:
## Claude review
**Verdict:** LGTM with nits
**Scope:** Replaces two split AI-suggestion popovers with a unified selection-based review surface, adds run provenance snapshotting into `extraction_runs.results` JSONB (no migration), and converts silent no-information abstentions into first-class `{value: null}` proposals — establishing transparency/traceability as a constitution pillar.

### Findings

- **medium** `backend/app/services/section_extraction_service.py` ~`:459` (batch path) — The plan's R3 calls out three `complete_run` sites (single, batch, per-section). The diff shows `_run_provenance` merged at two sites; the batch `complete_run` path was not visible in the diff. If batch-run proposals reference the batch `run_id` but provenance is never written to that run's `results`, those suggestions will silently return `null` provenance after every fresh run with no way to fix retroactively. **Verify the batch `complete_run` call also includes `results={"provenance": self._run_provenance}` (or equivalent merge).** If confirmed covered, this is a no-op nit.

- **medium** `backend/app/services/section_extraction_service.py` `_extract_with_llm` — `self._run_provenance = self._build_run_provenance(model, prompt_module, ...)` is set once after `extract_structured` is called. If there is any code path inside `_extract_with_llm` that iterates over multiple field groups or prompt modules (different `NAME`/`VERSION`), only the last iteration's params are snapshotted — the provenance would misattribute earlier fields' extraction to the final prompt. **Confirm `_extract_with_llm` either makes a single `extract_structured` call per invocation or that all calls within it are homogeneous (same prompt module, same model).** If the function is always single-call, this is fine.

- **low** `frontend/hooks/useAISuggestions.ts` `selectProposalCore` — After `setSuggestions(...)` updates the coord's `id/value/confidence` optimistically, the `.catch` path clears loading and shows a toast but does not revert the in-memory state. The coord will reflect the chosen (wrong) version until the next full cache invalidation. This pattern pre-exists in `acceptSuggestionCore`, but `selectSuggestion` from the review popover now exercises it on version-switches too (not just quick-accept). Consider capturing the previous suggestion value before the optimistic set and restoring it in `.catch`.

- **low** `frontend/components/extraction/RunProvenanceDisclosure.tsx` `genericKeys` — Any key added to `extraction_runs.results["provenance"]` in future backend changes will auto-render verbatim in the reviewer UI (the `REGISTRY_KEYS` / `SUPPRESSED_KEYS` fallback path). This is a correct forward-compat trade-off per the design, but it should be documented as a team convention: `_build_run_provenance` is the sole write point, and any new provenance field must be reviewed for reviewer-visibility before adding (e.g. internal trace IDs, future A/B variant flags, or anything that might later carry sensitive data).

### Incident-class sweep

| Class | Status |
|---|---|
| BOLA | ok — `_load_run_provenance` operates on run ids already scoped by the caller's membership-gated proposal set; `selectSuggestion` backed by existing server-side proposal-triplet validation |
| TOCTOU | ok — provenance writes target JSONB `results` column (no state-machine transition); `record_proposal` is idempotent per implementation |
| Swallowed errors | finding (low) — optimistic state not rolled back on `selectProposalCore` backend failure (pre-existing, wider path now) |
| ApiResponse envelope | ok — `provenance` added to existing `AISuggestionItem` / `AISuggestionHistoryItem` schemas; TypeScript types regenerated |
| TanStack cache | ok — `selectSuggestion` delegates to the same `acceptSuggestion` service call as the existing inline-accept flow, inheriting the same invalidation chain |
| Schema drift | ok — `dict[str, Any] \| None` (Pydantic) ↔ `{[key: string]: unknown} \| null` (TypeScript); types regenerated; no-info `{"value": None}` proposal is backward-compatible with existing JSONB column |
| RLS | n/a — no new tables or policies; provenance lives in existing `extraction_runs.results` JSONB |

---

PR #448, opened 2026-06-30
fix(extraction): AI-suggestion review — keep inline value, entailment value labels, run provenance

raphaelfh:
## Why

Three reviewer-reported defects in the AI-suggestion review surface (extraction HITL):
1. the inline suggestion below a field **disappeared** after running AI extraction a second time;
2. cited evidence showed **"Not supported"** even for reasonable suggestions on coded select/boolean fields;
3. the **"How this was generated"** provenance metadata never appeared in the review popover.

## What changed

**1 — inline suggestion stays visible + better affordances**
- `load_suggestions` dedup now prefers the most recent **value-bearing** proposal so a later "no information" (`{value:null}`, recorded since #443) can't bury an earlier real value; falls back to the latest no-info only when no run found a value. [extraction_suggestion_read_service.py](backend/app/services/extraction_suggestion_read_service.py)
- The inline value/confidence (and the no-info indicator) now open the **same** review popover as the History icon; both open `align="end"` so they clear the PDF/markdown panel. [AISuggestionDisplay.tsx](frontend/components/extraction/ai/AISuggestionDisplay.tsx), [FieldInput.tsx](frontend/components/extraction/FieldInput.tsx)

**2 — entailment judges the human label, not the raw code**
- The entailment judge was handed the select option **code** (`"Y"`) instead of its **label** (`"Yes"`), which a prose quote can't entail → systematic `unsupported`. New pure helper `value_str_for_claim` resolves select/multiselect/boolean values to their label before the claim; numeric/date/text pass through (deterministic numeric check preserved). [claim_value.py](backend/app/llm/claim_value.py), [section_extraction_service.py](backend/app/services/section_extraction_service.py)
- Added a clarifying tooltip on the attribution badge: it grades the cited **quote**, independent of the AI's confidence/rationale. [AISuggestionEvidence.tsx](frontend/components/extraction/ai/AISuggestionEvidence.tsx)

**3 — run provenance persists for session runs**
- Session-run extractions never wrote `results['provenance']` (`extract_section` skipped it when a `run_id` was passed; `extract_for_run` omitted it), so the disclosure had no data. New repo method `merge_results` persists provenance into the `results` JSONB **without** completing the still-live session run; both paths now write it. [extraction_run_repository.py](backend/app/repositories/extraction_run_repository.py), [section_extraction_service.py](backend/app/services/section_extraction_service.py)

**Cleanups (/simplify):** shared `ATTRIBUTION_COPY` table; hoisted `FieldInput` review binding; `{...review}` spread; shared `claim_value.normalize_options` reused by `schema._enum_values`; English-only docstrings; file-size baseline bump for two intentionally-grown files.

## Risk

- **Dedup semantics change** (read-only): the inline strip prefers latest-with-value; full history (incl. abstentions) is unchanged and still in the review popover. Covered by 2 new integration tests (mixed + all-no-info).
- **`merge_results`** is a read-modify-write on `results` JSONB but only ever merges the `provenance` key and preserves run status (not a state transition), so no run-state TOCTOU. New integration test asserts the session run stays in EXTRACT.
- No schema/endpoint/migration/RLS changes. Both affected endpoints remain BOLA-gated (`ensure_project_member`).
- **Forward-only:** runs extracted before this deploys keep empty provenance until re-extracted (FE degrades gracefully).

## Test plan

- [x] `make lint-backend` clean
- [x] `make test-backend` passes — **2308 passed** (5 new tests: dedup ×2, provenance ×1; value→label unit ×12; FE tooltip/inline already counted below)
- [x] `npm run lint` clean
- [x] `npm run typecheck` clean
- [x] `npm run test:run` passes — **929 passed**

## Out of scope (follow-ups)

- Resolve the option code→label for the **human-facing** suggestion card too (today it still shows `"Y"`); needs a FE label map or carrying the label in `proposed_value`.
- Consolidate provenance persistence into a single choke-point (`_create_suggestions`); requires changing `complete_run` merge-vs-replace semantics (also used by `model_extraction_service`).
- `_build_premise` robustness for multi-block/elided quotes (already spawned as a separate task).

### DIFF
--- a/backend/app/llm/claim_value.py
+++ b/backend/app/llm/claim_value.py
@@ -1,0 +1,78 @@
+"""Human-readable value rendering for the entailment CLAIM.
+
+A select/multiselect field stores the option CODE (e.g. "Y"), because the
+extraction output schema constrains the LLM to ``opt["value"]`` (see
+``app.llm.schema._enum_values`` / ``_value_type``). Handing that bare code to
+the entailment judge as ``CLAIM: "<field> = Y"`` is near-uninterpretable
+against prose, biasing coded select/boolean fields toward ``weak`` /
+``unsupported`` regardless of the citation quality. Resolve the code to its
+human label ("Yes") before building the claim so the judge can actually assess
+entailment.
+
+Numeric / date / text values pass through unchanged so the deterministic
+numeric check (``app.llm.value_support.numeric_value_supported``) still sees the
+raw number. Pure, no IO.
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+
+def normalize_options(allowed_values: Any) -> list[Any]:
+    """Return the raw option list from an ``allowed_values`` payload.
+
+    Tolerates both shapes the template builder emits: ``{"options": [...]}`` or a
+    bare ``[...]``; anything else yields ``[]``. Single source of the shape
+    tolerance, shared by ``option_label_map`` here and
+    ``app.llm.schema._enum_values`` so a new option encoding is handled once.
+    """
+    if isinstance(allowed_values, dict) and "options" in allowed_values:
+        options = allowed_values["options"]
+    elif isinstance(allowed_values, list):
+        options = allowed_values
+    else:
+        return []
+    return list(options or [])
+
+
+def option_label_map(allowed_values: Any) -> dict[str, str]:
+    """Map each select option's stored value (code) to its human label.
+
+    Options are ``{"value","label"}`` dicts or plain strings. Plain strings (and
+    options without a distinct label) map to themselves, so resolution is a safe
+    no-op for label-less templates.
+    """
+    out: dict[str, str] = {}
+    for opt in normalize_options(allowed_values):
+        if isinstance(opt, dict) and "value" in opt:
+            label = opt.get("label")
+            out[str(opt["value"])] = str(label) if label else str(opt["value"])
+        elif isinstance(opt, str):
+            out[opt] = opt
+    return out
+
+
+def value_str_for_claim(*, field_type: str | None, allowed_values: Any, value: Any) -> str:
+    """Render *value* as a human-readable string for the entailment CLAIM.
+
+    - ``bool``            -> "Yes" / "No"
+    - ``select``          -> the option's label for the code (fallback: the code)
+    - ``multiselect``     -> comma-joined option labels (fallback: each code)
+    - everything else     -> ``str(value)`` unchanged (numeric / date / text keep
+                             their raw form so the deterministic numeric check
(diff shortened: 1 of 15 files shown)


raphaelfh:
## Claude review
**Verdict:** LGTM with nits
**Scope:** Three targeted bug fixes in the HITL AI-suggestion surface: (1) dedup now prefers the most recent value-bearing proposal so a later no-info abstention can't blank an earlier real suggestion; (2) entailment judge now receives the select/boolean option's human label instead of the raw code, fixing systematic "Not supported" on coded fields; (3) session-run extractions now persist `results['provenance']` via a new `merge_results` repo method so the "How this was generated" popover has data. No schema, endpoint, migration, or RLS changes.

### Findings

- **nit** `backend/app/repositories/extraction_run_repository.py:152` — `merge_results` is a read-modify-write: it `SELECT`s `run.results`, merges in Python, then `flush()`es. Two concurrent requests on the same run (e.g. a user double-clicking the AI button) would both read the stale `results`, and the last flush wins, potentially clobbering the first write's keys. For the current sole caller (`provenance` is idempotent and both writes would be identical) this is harmless in practice, but the method's public signature invites future callers where the race is real. Suggested fix: replace the read-merge-write with a single SQL-level merge — `UPDATE extraction_runs SET results = COALESCE(results, '{}') || :patch WHERE id = :run_id RETURNING *` — which is atomic on PostgreSQL's MVCC and removes the SELECT round-trip.

- **nit** `backend/app/services/section_extraction_service.py:538` — `extract_for_run` now includes `"provenance": run_provenance` in the `results` dict passed to `complete_run`, where `run_provenance` can be `None`. Previously the key was absent; now it's present as `null` in JSONB. The FE degrades gracefully, but any caller doing `"provenance" in results` (key-existence check) will now find `True` even when provenance data is missing. Low risk given the FE already guards against nulls, but worth being aware if the key's presence is ever used as a sentinel.

### Incident-class sweep

| Class | Status |
|---|---|
| BOLA | ok — no new endpoints; both affected endpoints remain behind `ensure_project_member` per PR body |
| TOCTOU | nit — `merge_results` is a Python-level read-modify-write (see finding above); no run-state machine changes |
| Swallowed errors | ok — `merge_results` errors propagate; `if provenance is not None` guard is correct |
| Envelope (ApiResponse) | n/a — no new endpoints |
| TanStack cache | n/a — purely UI + backend service changes; no new cache keys |
| Schema drift | ok — no SQLAlchemy/Pydantic/TypeScript schema changes |
| RLS | n/a — no new tables, no migrations |

> **Note:** `mergeable_state` is `dirty` (conflicts with `dev`) — needs a rebase before merge.

---

PR #449, opened 2026-06-30
Promote dev → main: extraction AI-suggestion review fixes (#448, #447)

raphaelfh:
## Promote dev → main (production)

Ships two extraction AI-suggestion fixes to production:

- **#448** — AI-suggestion review: keep the inline suggestion visible after a re-run (dedup prefers latest-with-value), resolve coded select values to their human label before the entailment judge (+ clarifying badge tooltip), and persist run provenance for session runs so "How this was generated" renders.
- **#447** — entailment premise spans multi-block citations via anchor block_ids.

## Verification

Both already merged green to `dev` (full CI). Merged-tree local gates re-run before the dev PR landed: backend 2311 passed, frontend 936 passed, lint/typecheck clean.

## Notes

- Merge-commit promotion (not squash/ff). Railway deploys from `main`; Post-deploy Smoke gates prod health. Supabase Preview is expected to fail on promotions (ignore).

### DIFF
--- a/backend/app/llm/claim_value.py
+++ b/backend/app/llm/claim_value.py
@@ -1,0 +1,78 @@
+"""Human-readable value rendering for the entailment CLAIM.
+
+A select/multiselect field stores the option CODE (e.g. "Y"), because the
+extraction output schema constrains the LLM to ``opt["value"]`` (see
+``app.llm.schema._enum_values`` / ``_value_type``). Handing that bare code to
+the entailment judge as ``CLAIM: "<field> = Y"`` is near-uninterpretable
+against prose, biasing coded select/boolean fields toward ``weak`` /
+``unsupported`` regardless of the citation quality. Resolve the code to its
+human label ("Yes") before building the claim so the judge can actually assess
+entailment.
+
+Numeric / date / text values pass through unchanged so the deterministic
+numeric check (``app.llm.value_support.numeric_value_supported``) still sees the
+raw number. Pure, no IO.
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+
+def normalize_options(allowed_values: Any) -> list[Any]:
+    """Return the raw option list from an ``allowed_values`` payload.
+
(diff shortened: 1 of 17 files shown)


raphaelfh:
## Claude review
**Verdict:** LGTM with nits
**Scope:** Squash-merge promotion (dev → main) of three extraction AI-suggestion fixes — (1) inline dedup now prefers the latest value-bearing proposal over a later "no information" abstention; (2) `_build_premise` uses `anchor.block_ids` to span multi-block citations instead of a single char-range containment scan; (3) select/boolean option codes resolved to human labels before the entailment judge claim is built — plus wiring the inline suggestion strip as a second entry point into the review popover, and attribution badges now carry tooltip copy explaining they grade the cited quote (not the AI confidence).

### Findings

- **nit** `backend/app/repositories/extraction_run_repository.py` — `merge_results` uses a read-modify-write pattern (`get_by_id` → mutate in Python → `flush()`) that is not atomic. Two concurrent `extract_section` calls on the same `run_id` (different sections in one session) could race; last flush wins and could overwrite the other section's provenance patch. In practice this is benign — the `provenance` key captures the model/provider/params for the whole run, so concurrent writes would produce identical data — but an atomic `UPDATE extraction_runs SET results = COALESCE(results, '{}'::jsonb) || :patch WHERE id = :run_id` (or `func.jsonb_set`) would eliminate the race entirely at negligible extra cost.

### Incident-class sweep

| Class | Status |
|---|---|
| BOLA | n/a — no new or modified endpoints |
| Run-state TOCTOU | nit — `merge_results` read-modify-write (see above); lifecycle transitions use `RunLifecycleService` correctly throughout |
| Swallowed errors | ok — no bare `except`, no `.catch(() => success)` patterns |
| ApiResponse envelope | n/a — no endpoint changes |
| TanStack cache | n/a — no queries or mutations added; no cache-key changes needed |
| Schema drift | ok — no model / Pydantic / TS type changes; `results` is untyped JSONB |
| RLS | n/a — no new tables or policy changes |

---

PR #453, opened 2026-06-30
fix(extraction): persist run provenance at the proposal choke-point

raphaelfh:
## Why

Run provenance (how an AI run's suggestions were generated) was written at four separate sites in `section_extraction_service`. A new suggestion-owning path would silently omit it — a bug that already shipped twice (#443, #448). Deferred from #448's `/code-review`.

## What changed

- **Choke-point**: `_create_suggestions` (the single place AI proposals are recorded) now persists the run `provenance` snapshot via `ExtractionRunRepository.merge_results`, so every proposal-owning run carries it — including any future path. Skipped when no LLM ran.
- **`complete_run` now MERGES `results`** instead of REPLACE, so the provenance the choke-point wrote survives completion. Verified safe for all callers: `model_extraction_service` and the batch primary run start from empty `results` (merge == replace); the section paths want their provenance preserved. Its docstring is translated to English while here.
- Removed the now-redundant explicit provenance writes from the single-section paths (`extract_section` both branches, `_extract_section_with_memory`). `extract_for_run` keeps its run-aggregate token write (its token-shape fix is a separate PR).

## Risk

`complete_run` semantics change (REPLACE → MERGE) on a core repository method. Audited every caller (`grep complete_run app/`): all five either start from `{}` (model extraction, batch primary run) or want the merged key preserved (section paths). The choke-point write adds one `merge_results` per proposal batch (a cached `get_by_id` + flush within the same transaction).

## Test plan

- [x] `make lint-backend` clean (ruff check + format)
- [x] `uv run pytest tests/unit/test_extraction_run_repository.py tests/unit/test_section_extraction_service.py tests/unit/test_model_extraction_service.py …` — 90 passed, incl. new `complete_run` merge-semantics (preserve prior keys, empty==replace, overwrite-on-conflict, missing-run) and choke-point provenance (persists when LLM ran, skips otherwise)
- [x] `uv run pytest tests/integration/test_section_extraction_gate.py tests/integration/test_section_extraction_evidence.py` — 8 passed (incl. the session-run provenance regression test, still green via the choke-point)
- [x] Architectural Fitness (layered-arch OK; file-size baseline bumped +9 for the choke-point, surgical 1-line change)

## Out of scope

P4 (provenance token-shape consistency for `extract_for_run`) and P5 (docstring sweep of the remaining repository methods) ship as separate PRs.

### DIFF
--- a/backend/app/repositories/extraction_run_repository.py
+++ b/backend/app/repositories/extraction_run_repository.py
@@ -102,26 +102,29 @@ class ExtractionRunRepository(BaseRepository[ExtractionRun]):
         run_id: UUID,
         results: dict[str, Any],
     ) -> ExtractionRun | None:
-        """
-        Marca uma execucao como concluida.
+        """Mark a run COMPLETED, shallow-merging *results* into its ``results`` JSONB.
+
+        MERGE (not REPLACE) so run-level data already recorded — notably the
+        ``provenance`` snapshot the proposal choke-point (``_create_suggestions``
+        → ``merge_results``) writes — survives completion. A REPLACE would clobber
+        it. Callers whose run still has empty ``results`` (model extraction, the
+        batch primary run) are unaffected: merging into ``{}`` equals a replace.
 
         Args:
-            run_id: execucao.
-            results: Resultados da execucao.
+            run_id: the run to complete.
+            results: top-level keys to merge into ``results``.
 
         Returns:
-            ExtractionRun atualizado or None.
+            The updated ExtractionRun, or None if the run does not exist.
         """
         query_start = perf_counter()
-        await self.db.execute(
-            update(ExtractionRun)
-            .where(ExtractionRun.id == run_id)
-            .values(
-                status=ExtractionRunStatus.COMPLETED.value,
-                completed_at=datetime.now(UTC),
-                results=results,
-            )
-        )
+        run = await self.get_by_id(run_id)
+        if run is None:
+            return None
+        run.status = ExtractionRunStatus.COMPLETED.value
+        run.completed_at = datetime.now(UTC)
+        # Reassign (not in-place mutate) so SQLAlchemy tracks the JSONB change.
+        run.results = {**(run.results or {}), **results}
         await self.db.flush()
         query_duration_ms = (perf_counter() - query_start) * 1000
         logger.info(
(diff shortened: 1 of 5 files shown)


raphaelfh:
## Claude review
**Verdict:** LGTM with nits
**Scope:** Consolidates run provenance persistence from four scattered call sites into a single choke-point (`_create_suggestions`), and changes `complete_run` from a REPLACE to a MERGE so prior keys (provenance) survive completion — fixes a recurring bug class where new extraction paths silently omit provenance.

### Findings

- **nit** `backend/app/repositories/extraction_run_repository.py:107–115` — `complete_run` was previously an atomic `UPDATE … SET results = :results` (one round-trip, no-op on missing run). It is now a read-modify-write: `get_by_id` → Python dict merge → `flush`. Two concurrent callers on the same run can both read the original `results`, merge independently, and the later flush silently overwrites the earlier one's merged keys (the old code had the same last-write-wins problem, just at the SQL layer). The PR description audits all five callers and finds them serialized in practice — that assumption is load-bearing. Suggested fix: add a comment in the method body naming the serialization assumption (e.g. "callers are serialized per run; concurrent completion is not expected"), or guard with a conditional `UPDATE … WHERE status != COMPLETED RETURNING *` if this ever moves to a fan-out context.

- **nit** `backend/app/services/section_extraction_service.py:1568` (new block) — `_create_suggestions` persists provenance only `if count and provenance is not None`. Before this PR, both `extract_section` and `_extract_section_with_memory` wrote provenance unconditionally via `complete_run` — so a run where the LLM ran but produced 0 suggestions (empty field result, no matching instances) would carry provenance. After this PR it won't. The stated goal is "every suggestion-*owning* run", which is consistent with the `count` guard, but the test `test_skips_provenance_when_no_llm_ran` only covers `_run_provenance = None` — not the case where `_run_provenance` is set but `count == 0`. Suggested fix: add a test for that edge case and update the `usage` param docstring from "None (or no LLM run) skips provenance" to "provenance is skipped when no LLM ran *or* when no suggestions were created" so future readers don't have to rediscover the gate logic.

### Incident-class sweep

| Class | Status |
|---|---|
| BOLA | n/a — no new endpoints or data-access paths |
| TOCTOU | nit — `complete_run` lost its atomic `UPDATE` guard; callers are audited as serialized |
| Swallowed errors | ok — `merge_results` exceptions propagate unchanged |
| ApiResponse envelope | n/a — no API surface changes |
| Schema drift | ok — no SQLAlchemy/Pydantic/TS changes; no migration needed |
| RLS | n/a — no schema or migration changes |

---

PR #454, opened 2026-06-30
refactor(extraction): unify the value-emptiness predicate

raphaelfh:
## Why

`extraction_suggestion_read_service._is_no_info` and `run_lifecycle_service._is_value_filled` were two copies of one rule — the frontend `isNoInfoValue` (`value === null | undefined | ''`). Deferred from #448's `/simplify`.

## What changed

- New pure module `app/services/value_semantics.py`: `unwrap_value_envelope` + `is_value_empty` + `is_value_filled` (exact inverses). No DB, no IO — layer-legal.
- `run_lifecycle_service` (completeness gate) imports `is_value_filled`; `extraction_suggestion_read_service` (suggestion dedup) imports `is_value_empty`. Both private copies removed.
- `value_envelope.py` (export/openpyxl-coupled) was deliberately **not** chosen as the home — the predicate is pure semantics, not an export concern.

## Risk

`is_value_filled` powers the finalize completeness gate (the historical "40%/can't-submit" area). Semantics are preserved **exactly**: the shared predicate keeps the one-level `{"value": X}` unwrap, so only `None` and `""` are empty; whitespace/0/False/[] stay filled and the gate is never stricter than the form. The two predicates were already equivalent inverses on every input proposals/resolved-values actually carry (they diverged only on a non-envelope dict, which neither caller receives — pinned in the truth-table test).

## Test plan

- [x] `make lint-backend` clean (ruff)
- [x] `uv run pytest tests/unit/test_value_semantics.py` — 14 passed (truth table incl. the gate-critical `None`/`""`/`{"value":null}`/`0`/`False`/`[]`/double-wrapped-unit cases)
- [x] `uv run pytest tests/integration/test_run_lifecycle_service.py tests/integration/test_suggestion_read.py` — 60 passed (completeness gate + suggestion dedup unchanged)
- [x] Architectural Fitness (layered-arch OK; file-size unchanged)

## Out of scope

P4 (token-shape) and P5 (docstring sweep) ship as separate PRs.

### DIFF
--- a/backend/app/services/extraction_suggestion_read_service.py
+++ b/backend/app/services/extraction_suggestion_read_service.py
@@ -43,6 +43,7 @@ from app.schemas.extraction_suggestion import (
     AISuggestionsResponse,
     EvidenceResponse,
 )
+from app.services.value_semantics import is_value_empty
 
 
 def _extract_block_ids(ev: ExtractionEvidence) -> list[int]:
@@ -79,19 +80,6 @@ def _resolve_status(decision: str | None) -> str:
     if decision == ExtractionReviewerDecisionType.REJECT.value:
         return "rejected"
     return "accepted"
-
-
-def _is_no_info(proposed_value: Any) -> bool:
-    """True when a proposal carries no actionable value (the model abstained).
-
-    Mirrors the frontend ``isNoInfoValue``: a bare ``None``, the JSONB envelope
-    ``{"value": None}`` (the shape recorded for a "no information" outcome since
-    #443), or an empty string all mean "no information".
-    """
-    if proposed_value is None:
-        return True
-    value = proposed_value.get("value") if isinstance(proposed_value, dict) else proposed_value
-    return value is None or value == ""
 
 
 async def _load_run_provenance(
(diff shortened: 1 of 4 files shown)


raphaelfh:
## Claude review
**Verdict:** LGTM
**Scope:** Extracts the shared value-emptiness predicate from two private copies (`_is_no_info` in the suggestion dedup service and `_is_value_filled` in the lifecycle completeness gate) into a single pure module `value_semantics.py`, with a 14-case truth-table test covering the gate-critical inputs.

### Findings

One thing worth noting (not a finding, already documented): `_is_no_info` used `.get("value")` on any dict, so it would return `True` (empty) for a non-envelope dict like `{"unit": "mg"}`. The new `is_value_empty` returns `False` (filled) for that shape, aligning with what `_is_value_filled` already did. This is a deliberate and correct change; the PR body and `test_dict_without_value_key_counts_as_filled` both pin the behaviour. The author's claim that neither caller receives a non-envelope dict is plausible given how the AI extraction serialises values.

### Incident-class sweep

| Class | Status |
|---|---|
| TOCTOU | n/a — state-machine logic unchanged, only helpers swapped |
| Swallowed errors | n/a — pure predicates, no error handling |
| ApiResponse envelope | n/a — no response shape changes |
| TanStack cache | n/a — backend-only |
| Schema drift | n/a — no model or Pydantic changes |
| RLS | n/a — no migrations |

---

PR #461, opened 2026-07-01
Promote dev to main — absent_reason write-path + AI marker (ADR-0016 Phase 1)

raphaelfh:
Promotes ADR-0016 **Phase 1** (write-path contract + AI `no_information` marker, #460) to production.

Single commit ahead of main: `be8cae75`. No schema/migration change (JSONB payload only). dev CI green; local full backend suite (2364 passed) + FE tests + tsc/eslint/ruff/arch-fitness clean; adversarial code review over the diff found 0 defects. Preflight: Railway `/health` 200, no auth/storage migration drift; only non-green gate is the standing Supabase advisor baseline (0 ERROR, all pre-existing, zero DDL in this change) — override confirmed.

### DIFF
--- a/backend/app/services/exports/value_envelope.py
+++ b/backend/app/services/exports/value_envelope.py
@@ -18,9 +18,21 @@ from __future__ import annotations
 from typing import Any, Protocol, runtime_checkable
 
 from app.models.extraction import ExtractionFieldType
+from app.services.value_semantics import AbsentReason, value_absent_reason
 
 # An openpyxl-writable scalar. NEVER a dict, NEVER a list.
 ResolvedScalar = str | int | float | bool | None
+
+
+# Stable per-code labels for a coded absent_reason disposition (ADR-0016). A
+# marker resolves to one of these instead of leaking a dict-stringify into a
+# cell. Sourced from the AbsentReason enum so the codes can't drift; the Phase-4
+# export legend reuses these exact strings so cell and legend can't diverge.
+_ABSENT_REASON_LABELS: dict[str, str] = {
+    AbsentReason.NO_INFORMATION.value: "No information",
+    AbsentReason.NOT_APPLICABLE.value: "Not applicable",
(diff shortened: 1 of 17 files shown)


raphaelfh:
## Claude review
**Verdict:** LGTM with nits
**Scope:** Promotes ADR-0016 Phase 1 to main — AI `not_found` proposals now carry `{"value": null, "absent_reason": "no_information"}` instead of a bare null, closing the numeric/date "not reported" finalize-gate gap; `ambiguous` is split out as a genuine value (not absent); the export resolver and the frontend save path are updated to handle the marker end-to-end.

### Findings

- **nit** `backend/app/services/exports/value_envelope.py:70` — `_ABSENT_REASON_LABELS[reason]` would raise `KeyError` at export time if a new `AbsentReason` member is added without a matching label entry. `value_absent_reason` validates against `_ABSENT_REASON_CODES` (derived from the enum), so any new code would pass the guard and then miss the dict. The `test_every_absent_reason_code_has_a_label` test is the drift guard and will catch this in CI before it ships — but a module-load assertion would surface it immediately on import:
  ```python
  assert set(_ABSENT_REASON_LABELS) == {r.value for r in AbsentReason}, \
      "add a label to _ABSENT_REASON_LABELS for every AbsentReason code"
  ```
  Low urgency given the existing test.

### Incident-class sweep

| Class | Status |
|---|---|
| BOLA | n/a — no new endpoints |
| TOCTOU | ok — no new state transitions; proposals are richer JSONB, not state mutations |
| Swallowed errors | ok — no new catch blocks |
| Envelope (ApiResponse) | n/a — no response-shape changes |
| TanStack cache | ok — autosave uses `lastSavedByKeyRef`, not TanStack mutation; no invalidation gap |
| Schema drift | ok — marker lives in JSONB; FE/BE vocabulary aligned; `test_every_absent_reason_code_has_a_label` is the drift guard |
| RLS | n/a — no DDL |


---

PR #473, opened 2026-07-02
fix(qa): gate blind-reveal on extract stage + peers_revealed

raphaelfh:
## Why

Security-review 2026-07-02 finding 6: the QA screen computed `canReveal` as `manager && isBlindMode` only, so a blind manager on a consensus/finalized run still saw the Reveal button — a dead/misleading affordance, since consensus entry already auto-reveals run-scoped (ADR-0015). Not a leak (manager-authorized; ADR-0012 says blind mode is not a confidentiality boundary). The extraction screen already carries the correct guards.

## What changed

- `frontend/pages/QualityAssessmentFullScreen.tsx` — `canReveal` now mirrors `ExtractionFullScreen.tsx:452-456`: additionally requires `runDetail?.run.stage === "extract"` and `!runDetail.peers_revealed`.
- `frontend/test/QualityAssessmentFullScreen.test.tsx` — new "blind-reveal stage guards" describe: positive control (blind manager during extract sees Reveal), consensus + finalized show no Reveal, and `peers_revealed` hides it even during extract. All four fail on the old code except the control (witnessed RED before the fix).

## Risk

Low. Frontend-only visibility change of one header affordance; the backend `setManagerReviewVisibility` endpoint and permissions are untouched (server stays authoritative). Worst case is over-hiding: a blind manager on a pending run no longer sees Reveal until the run reaches extract — same behavior extraction has shipped with.

No doc change needed (no endpoint/table/ADR change; behavior now matches what ADR-0015's auto-reveal already documents).

## Test plan

- [x] `npm run test:run` — 169 files / 1103 tests pass (includes the 4 new screen tests)
- [x] `npm run typecheck` clean
- [x] `npm run lint` clean
- [ ] Backend suite n/a (no backend diff; CI runs it regardless)
- Manual: as a blind manager, open a QA run in consensus or finalized — no Reveal chip in the header; open one in extract — Reveal chip present.

## Out of scope

Extraction screen unchanged (already correct). Other findings from the 2026-07-02 run-header declutter review are tracked separately.

### DIFF
--- a/frontend/pages/QualityAssessmentFullScreen.tsx
+++ b/frontend/pages/QualityAssessmentFullScreen.tsx
@@ -362,8 +362,15 @@ export default function QualityAssessmentFullScreen() {
     return () => document.removeEventListener("keydown", onKey);
   }, []);
 
-  // Reveal: manager can un-blind QA reviewer identities for this project.
-  const canReveal = permissions.userRole === "manager" && permissions.isBlindMode;
+  // Reveal (the persistent project-toggle): offered only to a blind manager
+  // DURING extract, mirroring the extraction screen. Once the run reaches
+  // consensus the run-scoped auto-reveal covers it (ADR-0015), so the
+  // persistent toggle is no longer surfaced.
+  const canReveal =
+    permissions.userRole === "manager" &&
+    permissions.isBlindMode &&
+    runDetail?.run.stage === "extract" &&
+    !runDetail.peers_revealed;
   const onReveal = () => {
     void setManagerReviewVisibility(projectId ?? "", "quality_assessment", true)
       .then(() => permissions.refresh())
--- a/frontend/test/QualityAssessmentFullScreen.test.tsx
+++ b/frontend/test/QualityAssessmentFullScreen.test.tsx
@@ -456,6 +456,110 @@ describe("QualityAssessmentFullScreen", () => {
   });
 });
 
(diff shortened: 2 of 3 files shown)


raphaelfh:
## Claude review
**Verdict:** LGTM with nits
**Scope:** Frontend-only guard fix: `canReveal` on the QA screen now requires `stage === "extract" && !peers_revealed`, matching the already-correct extraction screen; four new unit tests cover the positive control and the three negative cases (consensus, finalized, peers_revealed).

### Findings

- **nit** `frontend/pages/QualityAssessmentFullScreen.tsx` — File-size baseline regression (blocking CI). The file grew from the baseline cap of 824 to 831 lines; `check_file_size.py` exits 1. The growth is intentional (7 extra lines for the multi-condition guard + comment), so the fix is to update the baseline: run `python scripts/fitness/check_file_size.py --update-baseline` locally and commit the updated `check_file_size.baseline`. No code change needed.

### Incident-class sweep

| Class | Status |
|---|---|
| BOLA | n/a — no new endpoints; server permissions unchanged; frontend is affordance-only |
| Swallowed errors | ok — `canReveal` short-circuits safely before `runDetail.peers_revealed`; `onReveal` error path unchanged |
| TanStack cache | n/a — no new queries or mutations |
| Schema drift | ok — `peers_revealed` read from top-level run view response; `typecheck` passes |
| RLS | n/a — no migrations |

---

PR #474, opened 2026-07-02
fix(runs): blind-gate reviewers_ready — no peer ids for blind callers

raphaelfh:
## Why

The run-view payload (`GET /runs/{id}/view`, and the QA session-open embed) shipped the raw `reviewers_ready` user-id list to every caller, `POST /runs/{id}/ready` echoed the same list, and the RLS SELECT on `extraction_reviewer_ready` allowed any project member to read peer ready rows via PostgREST. A blind reviewer could learn WHICH peers marked ready — participation metadata the blind contract (ADR-0012) should not expose by id. Found in the 2026-07-02 run-header declutter security review (finding 2). This PR closes **both layers in lockstep** (architecture doc §3 rule).

## What changed

**API layer (single home of the scrub)**

- `ExtractionReviewerReadyService.ready_summary_from` now takes `caller_id` + `include_peers`: `reviewers_ready` carries only the caller's own entry unless peers are revealed (backend/app/services/extraction_reviewer_ready_service.py). Counts stay aggregate — the N/M hint leaks no identity.
- `build_run_view` passes its already-computed `peers_revealed` as `include_peers`, so the gate reuses the single blind source and cannot drift from the row filter (backend/app/services/extraction_run_read_service.py:374).
- `POST /runs/{id}/ready` stays caller-scoped always (the response is the caller's toggle echo; an unblinded caller reads the full list via `/view`).
- Schema docstrings document the contract; API types regenerated (docstring-only diff). Response **shape** unchanged — the frontend self-check (`reviewers_ready.includes(currentUserId)`, `ExtractionFullScreen.tsx:1080`) works as-is because the caller's own entry is preserved.

**RLS layer (migration `0041_reviewer_ready_select_rls`)**

- Supersedes 0029's member-wide SELECT (whose "knowing someone is done leaks no values" rationale this review reversed). Mirrors the 0025 pattern exactly: own row OR `is_project_arbitrator` OR finalized run. INSERT/UPDATE unchanged (already self-scoped in 0029).
- Architecture doc: ReviewerReady entry documents the scrub + lockstep RLS; migration-head line bumped; head-pin test updated.

## Risk

Read-path scoping + one RLS policy replace; no table/column change, no write-path change. The backend reads `extraction_reviewer_ready` as `service_role` (RLS bypassed) and the frontend never reads it via PostgREST (verified), so 0041 is behavior-neutral for the app — it closes the devtools path only. Unblinded callers (consensus members, managers with reveal, arbitrator in consensus, finalized runs) still get the full list on `/view`; RLS deliberately keeps the manager-as-arbitrator carve-out per ADR-0012's API-stricter-than-RLS split. Migration verified offline (`--sql`) both directions and roundtripped in the suite.

## Test plan

- [x] New integration tests, written RED first:
  - blind reviewer's `/view` + `/ready` payloads contain no peer ids while `ready_count` stays aggregate; unblinded caller keeps the full list (`test_build_run_view.py`, `test_extraction_runs_ready_api.py`)
  - RLS probes as `authenticated` + JWT sub (`test_reviewer_ready_rls.py`): blind reviewer denied peer rows but keeps own; arbitrator sees all; finalized reveals; structural policy-expression guard
- [x] `make test-backend` — 2418 passed, 5 skipped (includes migration roundtrip through 0041)
- [x] `make lint-backend` clean
- [x] `npm run test:run` — 1099 passed; `npx tsc --noEmit` clean
- [x] `npm run generate:api-types` — committed, docstring-only diff

## Out of scope

None — the RLS follow-up originally chip-filed from this finding is folded in here so both layers land as one lockstep unit.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -235,14 +235,25 @@ jobs:
             role TEXT
           );
           -- Stub auth.uid() and auth.role(): real implementations live in
-          -- Supabase's auth schema. Alembic migrations reference these via
-          -- RLS policies, so we need them defined for `alembic upgrade head`
-          -- to succeed in CI. Returning NULL is fine for migration time —
-          -- tests don't exercise RLS via these helpers.
+          -- Supabase's auth schema. They mirror Supabase's readers of the
+          -- request.jwt.* GUCs so RLS probes (set_config claims + SET LOCAL
+          -- ROLE authenticated, e.g. test_reviewer_ready_rls.py) authenticate
+          -- exactly as against a real Supabase Postgres; with no claims set
+          -- they return NULL, which is all `alembic upgrade head` needs.
           CREATE OR REPLACE FUNCTION auth.uid() RETURNS uuid
-            LANGUAGE sql STABLE AS $$ SELECT NULL::uuid $$;
+            LANGUAGE sql STABLE AS $$
+              SELECT coalesce(
+                nullif(current_setting('request.jwt.claim.sub', true), ''),
+                nullif(current_setting('request.jwt.claims', true), '')::jsonb ->> 'sub'
+              )::uuid
+            $$;
           CREATE OR REPLACE FUNCTION auth.role() RETURNS text
-            LANGUAGE sql STABLE AS $$ SELECT NULL::text $$;
+            LANGUAGE sql STABLE AS $$
+              SELECT coalesce(
+                nullif(current_setting('request.jwt.claim.role', true), ''),
+                nullif(current_setting('request.jwt.claims', true), '')::jsonb ->> 'role'
+              )
+            $$;
           -- Supabase pre-creates these roles; migrations GRANT to them.
           DO $$ BEGIN
             IF NOT EXISTS (SELECT 1 FROM pg_roles WHERE rolname = 'anon') THEN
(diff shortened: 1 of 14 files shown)


raphaelfh:
## Claude review
**Verdict:** LGTM
**Scope:** Adds a blind-gate scrub to `reviewers_ready` by making `ExtractionReviewerReadyService.ready_summary_from` the single home of the filter — `include_peers=False` returns only the caller's own entry, while `build_run_view` feeds `include_peers=detail.peers_revealed` (the existing single blind source) so the scrub cannot drift from the row-filter on `/view`; `POST /ready` is hardwired to caller-only as a toggle echo.

### Findings

Key verifications performed:
- `useMarkReady.ts:27-29` `onSuccess` invalidates `runsKeys.detail` and `runsKeys.reviewers` without consuming the mutation response's `reviewers_ready`, so the POST /ready behavioral change (now always caller-scoped, even for managers) cannot regress the frontend self-check at `ExtractionFullScreen.tsx:1080`.
- `ExtractionFullScreen.tsx:1080`: `isReady` is derived from `runDetail?.reviewers_ready` (from GET /view, not the mutation response). Managers still see the full list on GET /view via `include_peers=detail.peers_revealed=true`. ✓
- `QualityAssessmentFullScreen.tsx` has no reference to `reviewers_ready` — QA screen is not affected. ✓
- The RLS residual gap on `extraction_reviewer_ready` (0029's project-wide SELECT) is correctly documented in the PR body and in the architecture doc with a follow-up task. The API layer is deliberately stricter than RLS pending that revisit. ✓

### Incident-class sweep

| Class | Status |
|---|---|
| BOLA | ok — no new endpoints; existing `_load_run_and_check_member` + `ensure_project_reviewer` gates on `POST /ready` unchanged; `GET /view` membership gate unchanged |
| Swallowed errors | ok — `[uid for uid in ready if uid == caller_id]` is a plain list comprehension; no silent error path |
| ApiResponse envelope | ok — `POST /ready` still returns `ApiResponse.success(RunReadyStateResponse(**summary))` |
| TanStack cache | ok — mutation `onSuccess` invalidates; response `reviewers_ready` is never written into cache state directly |
| Schema drift | ok — response shape unchanged (same fields); frontend types regenerated via `npm run generate:api-types`; semantics documented in type comments |
| RLS | ok (residual gap acknowledged in PR body + architecture doc; follow-up task filed; API is deliberately stricter than RLS) |

---

PR #479, opened 2026-07-03
fix(tests): run migration round-trips on an ephemeral scratch DB

raphaelfh:
## Why

The alembic round-trip tests (`backend/tests/integration/test_migration_roundtrip.py`) ran `alembic downgrade/upgrade` subprocesses against the **shared local dev database**. Two confirmed consequences:

1. **Permanent catalog pollution.** Every `make test-backend` left ~63 dropped-column tombstones in `pg_attribute` (`feedback_reports` +21/run from 0020's add/drop churn, `article_files` +16, `extraction_fields` +14, `extraction_evidence` +11). Postgres counts tombstones toward the 1600-column limit until the table is recreated — `VACUUM FULL` does **not** clear them (verified empirically). After ~75 runs `feedback_reports` hit the wall: 21 live columns + 1,579 tombstones = 1,600 → `TooManyColumnsError` on any migration touching it.
2. **Silent data destruction.** The downgrades wiped dev data on every run: 0033's downgrade drops/recreates `article_files.content_markdown` (all parsed markdown nulled), and the downgrade-to-0001 drops `article_text_blocks` entirely.

## What changed

- **`backend/tests/integration/supabase_stub.sql` (new)** — the CI auth/storage stub, extracted verbatim from the two duplicated inline heredocs in `ci.yml`. Single source of truth; idempotent.
- **`backend/tests/integration/test_migration_roundtrip.py`** — new session-scoped `migration_db_url` fixture: `DROP/CREATE DATABASE prumo_migration_test WITH (FORCE)` via an autocommit asyncpg admin connection, apply the stub, `alembic upgrade head` — exactly how CI builds its test DB. The 9 DDL round-trip tests + the head-pin test now assert through a function-scoped `migration_session` bound to the scratch DB. `_run_alembic` takes a required `database_url` kwarg and overrides **both** `DATABASE_URL` and `DIRECT_DATABASE_URL` in the subprocess env (alembic's `env.py` prefers `DIRECT_`; OS env beats `backend/.env`). Also fixes the stale module docstring that claimed a skip-when-unreachable mechanism that never existed, and applies the project's `sslmode→ssl` rename when building the asyncpg engine URL.
- **`.github/workflows/ci.yml`** — both stub heredocs (backend-test, backend-e2e) replaced with `psql … -f tests/integration/supabase_stub.sql`.
- **`test_migration_0039_backfill.py` untouched by design** — it needs the committed SEED graph and already runs the data-only migration SQL inside a rolled-back savepoint (zero tombstones).

## Risk

Test-infrastructure only; no app code, no migrations. The scratch DB is created on the same Postgres server the tests already use (CI service-container user and local Supabase `postgres` both have CREATEDB; verified locally with `rolcreatedb=t`). In CI the fixture now runs one extra `alembic upgrade head` (~seconds). Adversarially reviewed (3-dimension fan-out + refutation pass): one confirmed finding (`sslmode` query param would crash the scratch engine on pooler-shaped URLs) — fixed in this PR.

## Test plan

- [x] `make lint-backend` clean (530 files formatted, all checks passed)
- [x] `make test-backend` full suite: **2419 passed, 5 skipped** (same skip set as baseline) in 1m08s
- [x] Round-trip module alone: 11/11 in ~16s
- [x] Main-DB `pg_attribute` tombstone counts byte-identical before/after the full suite (previously +63/run)
- [x] Canary row planted in `article_files.content_markdown` survived the suite (previously nulled by 0033's round-trip); reverted after
- [x] `prumo_migration_test` absent from `pg_database` after the session (teardown drop works)

## Out of scope

- A future squash of the migration chain would reset the ~117 tombstones a single fresh `upgrade head` legitimately leaves (chain-internal add/drop churn) — harmless at current scale.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -219,67 +219,10 @@ jobs:
         run: bash ../scripts/validate_migration_boundaries.sh
         working-directory: backend
 
+      # Stub content lives in tests/integration/supabase_stub.sql — the same
+      # file the migration round-trip fixture applies to its scratch DB.
       - name: Bootstrap auth schema stub (Supabase auth.users + RLS helpers)
-        run: |
-          psql "${DATABASE_URL}" -v ON_ERROR_STOP=1 <<'SQL'
-          CREATE SCHEMA IF NOT EXISTS auth;
-          -- Supabase's auth.users carries far more columns (instance_id, aud,
-          -- role, encrypted_password, ...). Tests that exercise the JWT path
-          -- — notably test_membership_guards.outsider_user — insert with the
-          -- minimum useful subset, so mirror it in the stub.
-          CREATE TABLE IF NOT EXISTS auth.users (
-            id UUID PRIMARY KEY,
-            email TEXT,
-            instance_id UUID,
-            aud TEXT,
-            role TEXT
-          );
-          -- Stub auth.uid() and auth.role(): real implementations live in
-          -- Supabase's auth schema. They mirror Supabase's readers of the
-          -- request.jwt.* GUCs so RLS probes (set_config claims + SET LOCAL
-          -- ROLE authenticated, e.g. test_reviewer_ready_rls.py) authenticate
-          -- exactly as against a real Supabase Postgres; with no claims set
-          -- they return NULL, which is all `alembic upgrade head` needs.
-          CREATE OR REPLACE FUNCTION auth.uid() RETURNS uuid
-            LANGUAGE sql STABLE AS $$
-              SELECT coalesce(
-                nullif(current_setting('request.jwt.claim.sub', true), ''),
-                nullif(current_setting('request.jwt.claims', true), '')::jsonb ->> 'sub'
-              )::uuid
-            $$;
-          CREATE OR REPLACE FUNCTION auth.role() RETURNS text
-            LANGUAGE sql STABLE AS $$
-              SELECT coalesce(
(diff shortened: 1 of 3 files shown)


raphaelfh:
## Claude review
**Verdict:** LGTM
**Scope:** Extracts the duplicated CI inline SQL auth-stub heredoc into a single `supabase_stub.sql` source of truth, and refactors all migration round-trip tests to run against an ephemeral scratch DB (`prumo_migration_test`) rather than the shared dev DB — preventing `pg_attribute` tombstone accumulation and accidental dev data destruction from repeated downgrade/upgrade cycles.

### Findings

### Incident-class sweep

| Class | Status |
|---|---|
| Swallowed errors | ok — subprocess failure raises (`check=False` + explicit assert on returncode); asyncpg exceptions propagate |
| Envelope | n/a — no API changes |
| Schema drift | n/a — no model/Pydantic/TS changes |
| RLS | n/a — no new tables or policies |

**Notes:** Design decisions are sound throughout — `DROP DATABASE IF EXISTS ... WITH (FORCE)` at setup self-heals a prior crash; `NullPool` prevents pooled connections blocking the session-end `DROP DATABASE`; the `migration_db_url` fixture yields a plain `postgresql://` URL and `migration_session` explicitly sets `drivername="postgresql+asyncpg"` (the split is intentional); `test_migration_0039_backfill` intentionally stays on the main DB in a rolled-back savepoint (documented in the docstring). All stub SQL is idempotent.

---

PR #480, opened 2026-07-03
fix(templates): propagate config edits to article extraction/QA forms

raphaelfh:
## Why

Template configuration edits (new sections/fields, allow "Other (specify)", allow "Not applicable", allow "Not evaluated") never appeared in the per-article extraction/QA forms — not even for articles opened after the edit. Edits are written via PostgREST, nothing ever refreshed the active `extraction_template_versions.schema_`, and run views render from the run's frozen snapshot (`extraction_run_read_service._entity_types_for_run`). A second, latent landmine: the clone heal compared live structure to the **global** template, so any customized clone was one re-import / QA session open away from a destructive wipe.

## What changed

**Backend**
- `app/services/template_version_service.py` (new): `republish` freezes the live structure into a NEW active version (v+1; prior rows never mutated — `consensus`/`finalized` runs stay pinned), re-pins `pending`/`extract` runs, and materializes missing singleton instances via `HITLSessionService.ensure_instances` (otherwise the ADR-0009 finalize gate — per EXISTING instance — silently skips a newly added required section). Lock ordering mirrors session open: advisory `(article, template)` locks first (sorted), then the template row `FOR UPDATE`; ownership is validated with an unlocked read before any lock.
- `POST /projects/{project_id}/templates/{template_id}/republish-version` (manager-gated, 404 on cross-project ids) in `project_templates.py`.
- `run_lifecycle_service.create_run`/`reopen_run` take `FOR SHARE` on the template row so version resolution serializes with a concurrent republish (a run could otherwise pin the just-deactivated version and miss the re-pin UPDATE).
- `template_clone_service` heal: drift is now measured against the **active snapshot**; non-empty drift self-heals by **publishing live** (never wipe-and-rebuild — indistinguishable from a deliberate edit whose republish was lost, and the old wipe 500'd on the RESTRICT FK whenever instances existed). Zero-state clones still rebuild from the global. This also self-heals templates edited before this fix.

**Frontend**
- `RunViewFieldResponse` + `fieldFromRunView` now carry `allows_not_applicable`/`allows_not_evaluated` — the backend already serialized them; the hand-written adapter dropped them, so the NA/NE markers never rendered on the extraction page (QA reads live PostgREST structure, which is why ADR-0016 worked there).
- `useTemplateRepublish` (new) fires after every config mutation (field add/update/delete/reorder, other-specify, section add/remove/label, template import) — fire-and-forget after the success toast — and invalidates `runsKeys` + `templateEntityTypesKeys` (factory moved to `lib/query-keys/extraction.ts`).

**Docs**: `extraction-hitl-architecture.md` §2 (stage-aware pinning), clone-heal row, services list updated.

## Risk

- **Re-pin semantics change**: runs in `pending`/`extract` now follow template edits (this is the reported expectation); `consensus`/`finalized` stay frozen — version rows are still immutable.
- **Heal contract change**: bc055915-style corruption (live < snapshot, out-of-band loss) no longer factory-resets on re-clone; live is authoritative and the snapshot self-heals. Factory recovery = delete template + re-import. Pinned in `test_template_clone_service.py` with rationale.
- Known limitation (same as before for other writers): global seed updates no longer force-refresh existing clones (the old path 500'd on in-use templates anyway).

## Test plan

- [x] `backend: uv run pytest` — 2429 passed, 5 skipped (includes 9 new tests: republish v+1/no-op/re-pin/instances/BOLA/endpoint; heal self-heal ×2)
- [x] `npm run test:run` — 1126 passed (new: adapter disposition-flag test, useFieldManagement republish wiring ×4)
- [x] `npx tsc -p tsconfig.app.json --noEmit`, `npm run lint`, `ruff check` + `format --check` — clean
- [x] Fitness gates green (file-size baseline updated for the copy dictionary, +2 lines)
- [ ] Manual: edit a field's flags in Configuration → open the article form → new section/field/markers render

## Out of scope

- Moving section/field CRUD behind the API (read/write-path consolidation) — republish-on-edit is the interim contract.
- A backfill republish for templates edited before this fix: they self-heal on the next config edit, re-import, or QA session open.

### DIFF
--- a/backend/app/api/v1/endpoints/project_templates.py
+++ b/backend/app/api/v1/endpoints/project_templates.py
@@ -8,6 +8,11 @@
   by configuration flows that enable QA tools.
 * ``PATCH /api/v1/projects/{project_id}/templates/{template_id}`` — toggle
   ``is_active`` (e.g. disable a QA tool in Configuration).
+* ``POST /api/v1/projects/{project_id}/templates/{template_id}/republish-version``
+  — publish the live section/field structure as a new active
+  ``extraction_template_versions`` row. Called by the configuration UI after
+  every section/field edit (the edits themselves go through PostgREST), so
+  article forms — which render from the run's version snapshot — pick them up.
 
 These are project-scoped. ``POST /api/v1/hitl/sessions`` is per-article run
 lifecycle and is separate.
@@ -23,6 +28,7 @@ from app.schemas.common import ApiResponse
 from app.schemas.hitl_session import (
     CloneTemplateRequest,
     CloneTemplateResponse,
+    RepublishTemplateVersionResponse,
     TemplateKind,
     UpdateTemplateActiveRequest,
     UpdateTemplateActiveResponse,
@@ -36,6 +42,7 @@ from app.services.template_clone_service import (
     TemplateCloneService,
     TemplateNotFoundError,
 )
+from app.services.template_version_service import TemplateVersionService
 
 router = APIRouter()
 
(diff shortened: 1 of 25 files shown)


raphaelfh:
## Claude review
**Verdict:** LGTM with nits
**Scope:** Adds `TemplateVersionService.republish` (new backend service + endpoint) to freeze the live section/field structure into a new active version row after every config mutation, re-pins editable-stage runs to the new version, backfills singleton instances for newly added sections, and wires it into every frontend config mutation via `useTemplateRepublish`; also fixes `fieldFromRunView` dropping `allows_not_applicable`/`allows_not_evaluated` and changes clone-heal to self-heal the snapshot from live rather than wipe-and-rebuild.

### Findings

- **nit** `backend/app/services/template_clone_service.py:299` — Local import of `TemplateVersionService` avoids a circular dependency (`template_version_service.py` imports `TemplateNotFoundError` from this module). Functional and documented, but signals coupling. Clean-up: extract `TemplateNotFoundError` to `app/services/exceptions.py` in a follow-up so both modules import from a shared leaf.

- **nit** `frontend/hooks/extraction/useTemplateRepublish.ts:42` — `republish` is an `async const` defined inside the hook body and recreated on every render; no `useCallback`. No current bug (callers invoke it imperatively, never as a dependency), but wrap it in `useCallback([projectId, templateId])` before any refactor passes it as a prop.

### Incident-class sweep

| Class | Status |
|---|---|
| BOLA | ok — service-level `(project_id, project_template_id)` ownership check before any lock; manager gate on endpoint; cross-project BOLA tested at both service and endpoint layers |
| TOCTOU | ok — `create_run`/`reopen_run` take `FOR SHARE` on the template row so version resolution serialises with `republish`'s `FOR UPDATE`; advisory locks taken before `FOR UPDATE` inside `republish`, mirroring `open_or_resume` lock order; no inversion possible |
| Swallowed errors | ok — fire-and-forget is explicit (`void republish()`) with error toast on failure; no bare `except`; failed mutations do not call `republish` |
| Envelope | ok — endpoint returns `ApiResponse[RepublishTemplateVersionResponse]`; frontend uses `apiClient` (envelope-unwrapping layer) consistent with sibling endpoints |
| TanStack cache | ok — `runsKeys.all` + `templateEntityTypesKeys.byTemplate(templateId)` invalidated on success; import-path additionally invalidates `templateEntityTypesKeys.all` |
| Schema drift | ok — `RunViewFieldResponse` TS interface and `fieldFromRunView` adapter updated together; `openapi.json` and `schema.d.ts` regenerated; Pydantic schema unchanged (backend was already serialising these flags) |
| RLS | n/a — no new tables or migration; Supabase bot confirmed no `supabase/` changes |

---

